PR #1, opened 2017-01-17
Bugfixes

baszoetekouw:
Fixing some small generic bugs (see individual commits)

### DIFF
--- a/cmds.pl
+++ b/cmds.pl
@@ -174,7 +174,7 @@ sub task_set_priority {
     if ( $p eq 'N' ) {
       $p = '';
     }
-    my ($es,$result) = &task_exec("$id modify prio:$p");
+    my ($es,$result) = &task_exec("$id modify 'prio:$p'");
     if ( $es != 0 ) {
       $error_msg = $result;
       &draw_error_msg();
@@ -204,7 +204,7 @@ sub task_set_project {
     beep();
     return;
   }
-  my ($es,$result) = &task_exec("$id modify proj:$p");
+  my ($es,$result) = &task_exec("$id modify 'proj:$p'");
   if ( $es != 0 ) {
     $error_msg = $result;
     &draw_error_msg();
--- a/cmds.pl
+++ b/cmds.pl
@@ -10,8 +10,7 @@ sub prompt_quit {
     &draw_prompt_line('');
     return;
   }
-  endwin();
-  exit();
+  &clean_exit()
 }
 
 #------------------------------------------------------------------
--- a/cmds.pl
+++ b/cmds.pl
@@ -21,7 +21,8 @@ sub task_add {
     &draw_prompt_line('');
     return;
   }
-  # TODO: fix this evilness
+  # TODO: get rid off escaping by replacing the shell call in task_exec() by something like IPC::Open2().
+  #       That would allow us to get rid of all shell expansions, quotations and escaping. [BaZo]
   $str =~ s{[&^\\]}{\\\&}g;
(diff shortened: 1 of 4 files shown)


read.pl
```diff
@@ -128,16 +128,18 @@ sub inner_read_report {
...
-      $taskid2report[$i] = $prev_id;
+      audit("inner_read_report: report2taskid[$i]=$prev_id")
     }
     $prev_id = $report2taskid[$i];
     $i++;
   }
   close(IN);
 
   if ( $#report_tokens > -1 ) {
+    # WHY?!!!!
```
scottkosty: I have the same comment for a lot of VIT's code. Is your comment about why the if statement conditions on that particular condition, or is it about the line below the comment? For comments like these, let's use something more standard, like:

`# FIXME: I don't see why [insert reason] Either comment why or try removing? bz`

You do not need to sign (as I put above with bz), but I personally find it helpful.

cmds.pl
```diff
@@ -21,6 +21,8 @@ sub task_add {
     &draw_prompt_line('');
     return;
   }
+  # TODO: fix this evilness
```
scottkosty: From what I understand, escaping is all we can do, because we cannot quote the string because we want taskwarrior to parse it (e.g. the string might be "buy a cup of coffee +caffeine project:life").

Your comment suggests there's a more robust approach, but I don't see it. What do you suggest for the long-run?

read.pl
baszoetekouw: You're totally right.  Putting the comment like this is quite unhelpful, as I can't even rememeber now what is was about.  I had probably something to do with me trying to figure out what exactly the code is trying to accomplish here.

cmds.pl
baszoetekouw: The proper way to solve this would probably be to use IPC::Open2 in task_exec() instead of a regular open() with shell expansion.  Using open2() would allow you to avoid the shell altogether (with all its escaping and expansion pitfalls) and as a bonus allows you to specify each command line argument for ``task`` separately.

However, fixing this is quite some work, I think, as it would require rewriting not only the ``task_exec()`` function, but also large parts of ``cmds.pl``.

read.pl
```diff
@@ -125,19 +125,21 @@ sub inner_read_report {
       $i++;
       next;
     }
-    if ( $_ =~ /^\s*(\d+) / ) {
+    if ( $_ =~ /^\s{0,5}(\d+) / ) {
```
baszoetekouw: The proper way to fix this, I think, would be to force ``task`` to show each task on a single line (by setting ``defaultwidth`` to 0), and then handling our own line breaking and layout.  Of course that would be quite a departure from vit's current model, in which it simply shows whatever ``task`` outputs, so I'm not entirely sure if that is the way to go.

Alternatively, using a regexp here might be the wrong way to look at it.  What we really should be checking is if the first column is empty (in which case the line is a continuation of the previous task), or if is isn't (in which case this is a new task).  That shouldn't be too hard to implement.

baszoetekouw:
I've fixed the issues you raised, so I think this PR should be ok to merge now.

 I still need to marge these changes into the other pull requests, to avoid any conflicts there.

baszoetekouw:
If you like, I can change the exit code.  The main reason for the fix was vit not cleaning up the terminal properly on ctrl-C, but it should be straightforward to keep a non-zero exit code.

scottkosty:
I see. Thanks, Bas. Your commit works well -> merged.

PR #2, opened 2017-01-17
Feature+multilinetask

baszoetekouw:
Fix selection bar when a task is split over multiple lines

### DIFF
--- a/draw.pl
+++ b/draw.pl
@@ -80,7 +80,7 @@ sub draw_report_line {
   my ($x, $t, $cp, $str);
   $x = 0;
   my $task_is_selected = ( $taskid2report[$report2taskid[$i]]==$task_selected_idx );
-  audit(sprintf("draw_report_line: i=%d, r2t=%d, r2t2r=%d, selected=%d, is_selected=%d", 
+  &audit(sprintf("draw_report_line: i=%d, r2t=%d, r2t2r=%d, selected=%d, is_selected=%d", 
     $i, $report2taskid[$i], $taskid2report[$report2taskid[$i]], $task_selected_idx, $task_is_selected));
   if ( $mode eq 'with-selection' and $task_is_selected ) {
     $report_win->attron(COLOR_PAIR($COLOR_SELECTION));
--- a/draw.pl
+++ b/draw.pl
@@ -79,12 +79,15 @@ sub draw_report_line {
   my ($i,$line,$mode) = @_;
   my ($x, $t, $cp, $str);
   $x = 0;
-  if ( $mode eq 'with-selection' && $i == $task_selected_idx ) {
+  my $task_is_selected = ( $taskid2report[$report2taskid[$i]]==$task_selected_idx );
+  audit(sprintf("draw_report_line: i=%d, r2t=%d, r2t2r=%d, selected=%d, is_selected=%d", 
+    $i, $report2taskid[$i], $taskid2report[$report2taskid[$i]], $task_selected_idx, $task_is_selected));
+  if ( $mode eq 'with-selection' and $task_is_selected ) {
     $report_win->attron(COLOR_PAIR($COLOR_SELECTION));
     &set_attron($report_win,$selection_attrs);
   }
   for $t (0 .. $#{ $report_tokens[$i] } ) {
-    if ( $mode eq 'without-selection' || $i != $task_selected_idx ) {
+    if ( $mode eq 'without-selection'  or not $task_is_selected ) {
       my $fg = $report_colors_fg[$i][$t];
       my $bg = $report_colors_bg[$i][$t];
       $cp = &get_color_pair($fg,$bg);
@@ -95,7 +98,7 @@ sub draw_report_line {
     my $tok = $report_tokens[$i][$t];
     $report_win->addstr($line,$x,$tok);
     &set_attroff($report_win,$report_attrs[$i][$t]);
-    if ( $mode eq 'without-selection' || $i != $task_selected_idx ) {
+    if ( $mode eq 'without-selection' or not $task_is_selected ) {
       $report_win->attroff(COLOR_PAIR($cp));
     }
     $x += mbswidth($tok);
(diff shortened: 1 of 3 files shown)


scottkosty:
Thanks for this useful feature! I have been testing it and like it a lot. I think we need to be careful with making this the default behavior. My guess is that most VIT users would like this to be the default behavior, but I would like for the transition to be smooth.

First, let me explain why I think some (I would guess a minority) VIT users would prefer the current VIT behavior. If you select an annotation in VIT and press "D", it deletes the selected annotation. Further, I'm planning to introduce support for executing commands on annotations (e.g. with "o").  Although I would like to introduce it in an abstract way, I have taskopen <https://github.com/ValiValpas/taskopen> in mind. If taskopen is executed on a task with more than one annotation that matches a taskopen regex, taskopen asks which annotation you would like to process. Implementing behavior that allows you to run taskopen directly on a selected task would allow the user to skip this extra step.

I believe that most VIT users would prefer the behavior introduced by your patch. And that is why I would like to transition to that as the default.

Ideally, we would expose both behaviors simultaneously to users, so that they could easily switch between selecting tasks and selecting annotations. I wonder if we can rely on Vim for some inspiration. Using jk navigates logical lines in Vim, which is consistent with your patch. Currently, VIT behavior more like Vim's gj, gk, which navigate by screen line. I think this is similar to VIT's current behavior. So perhaps we could experiment with having your patch as the default behavior, and then if a user wants to select an annotation, they can do "gj" and "gk" to navigate among annotations. What do you think about this basic idea? We would want to think about the details, and probably trial and error is the only way to get something that feels right. For example, if the user is on the last annotation of a task, when they press "gj" I guess it should move to the first annotation of the next task?

I think we should consider two possible approaches:

Approach 1: introduce your patch as an opt-in setting (via vitrc) for VIT 1.3, and mention in our release notes that for the release of VIT after this one (that is, VIT 1.4), the option will be turned on by default.

Approach 2: implement the "gj", "gk" approach I outline above.

What are your thoughts?

I have been testing and found some minor bugs with your patch (or they expose bugs that were already present, but hidden, in VIT), which I list below. VIT has many such minor bugs already so I am in some sense being unfair in asking for a polished patch when VIT is far from polished, but hopefully we can eventually make VIT a bit more polished.

1. H does not work for me if the first line on the screen is an annotation.
Do you think the expected behavior is that VIT selects the first whole task that is shown or that it selects the task associated with the annotation? If the second, then a screen redraw would be necessary, which might feel a little jumpy to the user. But I really don't know what most users would expect.

2. If I scroll down a few tasks (just enough to force a redraw of the next tasks), and then back up to the start of my task list, the last task in my list is drawn as the first. Can you reproduce?

3. I have some problems when using search. For tasks with many (e.g. 15) annotations, when I get to them in the search results, they are not highlighted. Can you reproduce?

4. I came across unexpected behavior when scrolling up when on my screen the first line was of an annotation.
Suppose I have the following drawn on the screen:
--line 1 of my screen: second line of task 40
--line 2 of my screen: first line of task 62
If I currently have task 62 selected, if I press 'k', I get the following:
--line 1 of my screen: first line of task 15
--line 2 of my screen: second line of task 15
--line 3 of my screen: first line of task 40
--line 4 of my screen: second line of task 40
--line 5 of my screen: first line of task 62
task 40 was correctly selected, but I did not expect task 15 to be shown.

5. On a quick test of "filter" (using the "f" keybinding), the initial selection was not drawn. A simple 'j' or 'k' gets the selection back.

6. The help needs to be changed accordingly. For example, currently the help for "j" says "move down one line". We might want to wait on updating the help until we agree on what to do for 1.3.

Once we discuss a bit, I think the next step is to see if djp (IRC handle) is willing to join the discussion and help us test. Finally, we can take it to the users and get feedback from them. We don't have a mailing list, but I have the email address of several VIT users and testers. They are a valuable resource that we should involve in this feature.

Finally, let me give a personal thank you! Although I'm hesitant to change the default behavior of VIT for 1.3, I have no doubt that I will personally be using this feature. I don't use "D" to delete annotations.

scottkosty:
In testing, I discovered another reason why the current behavior of VIT is useful to me: the color of the task. Sometimes I want to know the color of the task that is selected. One way to get this is to move the selection away quickly (just a quick "j" or "k"), but I use a script where I process tasks one at a time in VIT (this helps me focus). And because there is always a selection, I cannot move it away.

Other than that, testing has gone well and I don't have any further bugs to report. I like it!

scottkosty:
Another comment: when marking a task as done, I think it would look better to flash the entire selection. Currently, only the "task part" (i.e. not the annotations) are flashed. I think this is done in
flash_current_task().

I just give this comment in case you agree and if it looks easy to fix. If not, then don't worry about it.

scottkosty:
If the last task in VIT's display has an annotation, if the task is selected and I press down again (which should be a noop), if I press up, nothing happens. On a second up, the selection correctly moves up a task. Can you reproduce?

baszoetekouw:
While testing further, I came across a nasty bug: the current patch assumes all tasks have a taskid.  But that's obviously not the case for completed or deleted tasks, so when trying to display those, vit will crash...
I think there is a way to work around this, but it requires changing the input parsing and lookup logic, so I need to spend some time getting this right...

scottkosty:
> While testing further, I came across a nasty bug: the current patch assumes all tasks have a taskid. But that's obviously not the case for completed or deleted tasks, so when trying to display those, vit will crash...

Ah yes, the ideal patch might be to use UUIDs internally, instead of relying on task IDs. Moving to UUIDs would help with a few issues in VIT. For example, see
https://bug.tasktools.org/browse/VT-17
https://bug.tasktools.org/browse/VT-14
If you thought it would help, I might spend some time (but not sure when) on moving from task IDs to UUIDs. Let me know.

> I think there is a way to work around this, but it requires changing the input parsing and lookup logic, so I need to spend some time getting this right...

OK. I'm thinking of moving forward with 1.3beta, and hopefully getting 1.3 out soon. This feature is useful enough that when it is ready we could go to beta quickly in the 1.4 cycle. Do you think that sounds like a good plan? I will look at the other pull requests. Do any of them depend on this one?

> While I'm at it, I should also be able to address the other issues you raise. I think definitely that being able to select individual annotations would be a nice feature; I should be able to add that without too much trouble.

Great!

baszoetekouw:
> Ah yes, the ideal patch might be to use UUIDs internally, instead of relying on task IDs. Moving to UUIDs > would help with a few issues in VIT. For example, see
> https://bug.tasktools.org/browse/VT-17
> https://bug.tasktools.org/browse/VT-14
> If you thought it would help, I might spend some time (but not sure when) on moving from task IDs to  UUIDs. Let me know.

That would make sense, actually.  I'm just not sure how much work it would be; taskwarrior doen't output the uuids by default after all, so we would need to change the output format of the reports for vit.  It might make more sense to use ```task export``` or something like that instead.  

> OK. I'm thinking of moving forward with 1.3beta, and hopefully getting 1.3 out soon. This feature is useful enough that when it is ready we could go to beta quickly in the 1.4 cycle. Do you think that sounds like a good plan? 

Yeah, that's fine.  That would give us some more time to think about how to deal with the issues mentioned above,  

> I will look at the other pull requests. Do any of them depend on this one?

I don't think there are any dependencies.

scottkosty:
> That would make sense, actually. I'm just not sure how much work it would be; taskwarrior doen't output the uuids by default after all, so we would need to change the output format of the reports for vit. It might make more sense to use task export or something like that instead.

I see. Yes this would take some time.

> I don't think there are any dependencies.

OK I'll look at them in the next few days.

scottkosty:
Note that this PR will fix #18.

thehunmonkgroup:

PR #3, opened 2017-01-17
add "wait" command

baszoetekouw:
Add command to issue a "wait" for a task

### DIFF
--- a/cmds.pl
+++ b/cmds.pl
@@ -218,6 +218,26 @@ sub task_set_project {
 
 #------------------------------------------------------------------
 
+sub task_set_wait {
+  my $id = $report2taskid[$task_selected_idx];
+  my $w = &prompt_str("Wait: ");
+  if ( $w eq '' ) {
+    &draw_prompt_line('');
+    return;
+  }
+  my ($es,$result) = &task_exec("$id modify 'wait:$w'");
+  if ( $es != 0 ) {
+    $error_msg = $result;
+    &draw_error_msg();
+    return;
+  }
+  $feedback_msg = "Modified task $id.";
+  &flash_current_task();
+  $reread_needed = 1;
+}
+
+#------------------------------------------------------------------
+
 sub shell_command {
   my $args = $_[0];
   my ($opts, $cmd);
--- a/getch.pl
+++ b/getch.pl
@@ -177,6 +177,11 @@ sub getch_loop {
         last CASE;
       }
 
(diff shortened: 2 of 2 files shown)


baszoetekouw:
This looks ok to merge; I just tested it on the latest 1.3 branch.

scottkosty:
@baszoetekouw I have a question about Git & Github workflow, but don't worry if you don't have time. It's a question I've had for a while though and I come across the situation often so I wonder what the recommended way to do it is.

The normal way to merge this branch would have been the following:

```
# Step 1: From your project repository, check out a new branch and test the changes.
git checkout -b baszoetekouw-feature+wait 1.3
git pull https://github.com/baszoetekouw/vit.git feature+wait

# Step 2: Merge the changes and update on GitHub.
git checkout 1.3
git merge --no-ff baszoetekouw-feature+wait
git push origin 1.3
```

When I did `git status` after the merge, it said that I was 5 commits ahead. The reason for this is because in your branch you did `git merge` to keep it up-to-date with respect to the 1.3 branch, which makes sense. But when I merge it, I wonder if it is cleaner to just merge the one commit in this case. Do you have an opinion on the best way to do it? I ended up doing the following:

```
git checkout -b baszoetekouw-feature+wait 1.3
git pull https://github.com/baszoetekouw/vit.git feature+wait
git checkout 1.3
git cherry-pick -x d8e23aec
```

Merging the first way has the advantage of keeping all the information. In some sense, all of the merges you did from upstream give an idea of when you implemented your patch and with respect to which commit of upstream 1.3 your patch was tested against. But in another sense, I find merging a merge of merges confusing.

baszoetekouw:
I myself mostly like to just merge te full tree (without "rewriting history", so to speak), or even just merge the PR on Github directly.  Otherwise, you might have to deal with resolving conflicts that have already been fixed in an earlier merge.

Another strategy that is often used, is a so-called "squash merge" (`git merge --squash`), which squashes the entire feature branch into a single diff and commits the diff as if it were a single commit.  The rationale there is that it keeps the upstream repo nice and clean (with the disadvantage of losing a bit of history).

In the end though, it all comes down to personal preference.

scottkosty:
Thanks, Bas. I think I will prefer to do a normal merge (preserving history) when that merge does not contain in it other merges (which can make the history hard to follow). In that case I will use the squash merge as you mention.

PR #4, opened 2017-01-17
add config option to disable pedantic prompts

baszoetekouw:


scottkosty:
Looks good. My only real comment/question:

What are your thoughts on renaming "pedantic" to "confirmation"? That would be
more consistent with .taskrc (see man taskrc).

Minor comments:

- I was going to fix the "code duplication" TODO, but realized it is not trivial because of the assignment.

- I think that all bools (i.e. $burndown) should be converted to use 1/undef, like you did for $pedantic. I will try to remember to do this.

- For the "TODO: get rid of the eval()", that would be nice! I agree it is a hack (which I introduced at 28450ae2) but I couldn't figure out how to do it in a better way.

baszoetekouw:
I agree dat "confirmation" is a bit more neutral ;)
I also documented the option in the man page.

Should be fine to go for 1.3, now.

baszoetekouw:
Regarding the eval(): are you sure it's even necessary at all?  the /e modifier in the regular expressions in replace_keycodes() seems to care care of the expansion of the KEY_F(1) &co?

scottkosty:
squash-merged at db23fc11.

scottkosty:
> Regarding the eval(): are you sure it's even necessary at all? the /e modifier in the regular expressions in replace_keycodes() seems to care care of the expansion of the KEY_F(1) &co?

No I'm not sure. If you're confident that we can remove it, let's go ahead and make a pull and we can do that for 1.3beta. If you're not confident, let's just leave it and we can experiment with a change after 1.3 is released.

PR #5, opened 2017-01-17
Feature+tags

baszoetekouw:
Add command to edit tags for tasks

### DIFF
--- a/cmds.pl
+++ b/cmds.pl
@@ -261,10 +261,22 @@ sub task_set_tag {
     return;
   }
 
-  foreach my $t (split(/\s\+/,$tags)) {
-    next if $t =~ m/^\s+$/;
+  # multiple tags can be specified separated by spaces
+  # keep track of the current modifier (default:+) and use it for subsequent tags
+  # so "+a b c" means "+a +b +c" and "+a -b c" means "+a -b -c"
+  my $mod='+';
+  foreach my $t (split(/\s+/,$tags)) {
+    next if $t =~ m/^\s*$/;
+
+    # check if a + or - was specified
     my $fc = substr($t,0,1);
-    $t = "+$t"  unless  $fc eq '+' or $fc eq '-';
+    if ( $fc eq '+'  or  $fc eq '-' ) {
+      # if so, save the current modifier
+      $mod = $fc;
+    } else {
+      # if not, add the current modifier
+      $t = $mod . $t;
+    }
 
     my ($es,$result) = &task_exec("$id modify '$t'");
     if ( $es != 0 ) {
--- a/cmds.pl
+++ b/cmds.pl
@@ -227,11 +227,6 @@ sub task_set_tag {
     &draw_prompt_line('');
     return;
   }
-  #my $tag = &task_info('Tag');
-  #if ( $p eq $proj ) {
-  #  beep();
-  #  return;
-  #}
 
   foreach my $t (split(/\s\+/,$tags)) {
     next if $t =~ m/^\s+$/;
--- a/cmds.pl
+++ b/cmds.pl
@@ -236,7 +236,7 @@ sub task_set_tag {
     my $fc = substr($t,0,1);
     $t = "+$t"  unless  $fc eq '+' or $fc eq '-';
 
-    my ($es,$result) = &task_exec("$id modify $t");
+    my ($es,$result) = &task_exec("$id modify '$t'");
     if ( $es != 0 ) {
       $error_msg = $result;
       &draw_error_msg();
--- a/cmds.pl
+++ b/cmds.pl
@@ -218,6 +218,39 @@ sub task_set_project {
 
 #------------------------------------------------------------------
 
+sub task_set_tag {
+  my $id = $report2taskid[$task_selected_idx];
+  my $tags = &prompt_str("Tag: ");
+  if ( $tags eq '' ) {
+    &draw_prompt_line('');
+    return;
+  }
+  #my $tag = &task_info('Tag');
+  #if ( $p eq $proj ) {
+  #  beep();
+  #  return;
+  #}
+
+  foreach my $t (split(/\s\+/,$tags)) {
+    next if $t =~ m/^\s+$/;
+    my $fc = substr($t,0,1);
+    $t = "+$t"  unless  $fc eq '+' or $fc eq '-';
+
+    my ($es,$result) = &task_exec("$id modify $t");
+    if ( $es != 0 ) {
+      $error_msg = $result;
+      &draw_error_msg();
+      return;
+    }
+  }
+
+  $feedback_msg = "Modified task $id.";
+  &flash_current_task();
+  $reread_needed = 1;
+}
+
+#------------------------------------------------------------------
+
 sub shell_command {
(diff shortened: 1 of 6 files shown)


scottkosty:
Nice work, Bas!

It is nice that the prompt history works well for this feature.

Here are some comments:

1. consider a tag called "mytag". If the task has the tag "mytag" and I would like to remove it, I can do the following keypresses:

  `T-myt<Tab><Enter>`

and that works as I expect. If I happen to do a second tab, the `-` is removed. Is there a reason for that? I see your comment "remove prefix + or - when matching tags". What is the purpose of that? Perhaps to signal to the user that a match was found?

2. if I do

  `Tone two<Enter>`

The *description* is modified to be "+one two". What do you think should happen in this case? I expected two tags to be added (+one +two), but perhaps that wasn't a good expectation.

3. I have a tag "computers". If I do

  `Tcompu<Tab>`

The tag is expanded to "computers", as expected. If I then press backspace until I remove the completion (e.g. I changed my mind about adding that tag), and then type

  `abcdef<Tab>`

It is expanded to "computers". Can you reproduce?

4. if I do T and then <tab>, I get:

  `----------`

which is not actually one of my tags. Can you reproduce? If not, let me know and I will take a look. It might have to do with my configuration. Did you consider parsing `task _tags` (note the underscore) instead of `task tags`? I think it would be faster, because `task tags` counts how often each tag is used, and is not as easy to parse. But I just checked `man task` and it says that `_tags` is deprecated in favor of:

  `task _unique tags`

But that is not quite what we want, since it shows all combinations of tags used (at least on 2.5.1 it does). I can ask the taskwarrior gurus on IRC if you agree that the parsing could be made more simple. But if you are happy with parsing `task tags`, I can be convinced to leave it as it is.

You don't have to address any of the following two comments for the branch to be merged. I just mention them in case you are motivated, and in case you have comments for if I want to take a look at them at a later time. If you don't have comments, please feel free to just ignore them.

5. For removing a tag, it would be nice to cycle only through the tags that the task currently has. I wonder how responsive this would be though, because we would have to query on the press of the tab.

6. It would be nice for `shift+tab` to cycle backwards. This is nice if you scroll quickly through tags with `<Tab>` and then overshoot by one.

Thanks for your work on this often-requested feature!

scottkosty:
Note also the typo "remove the tasg STRING" in the `commands` file.

baszoetekouw:
Hi Scott,

Thanks for testing this!  Issues 1, 2, 3, 4 and 6 should now be fixed.

I agree with you that for removing a tag, ideally vit would only cycle through the tags that are relevant for the current issue.  That would be pretty hairy to implement though, particularly in combination with supporting multiple tags in a single command.  I would therefor prefer the current solution, at least for now. 

Regarding the exact tag command to be used: I don;t think it really matters (at least, as long as speed is concerned).  Taskwarrior has to linearly parse the entire ~/.task/*.data files anyway to determine which tags are used; counting them while parsing should hardly give any overhead.

scottkosty:
> Regarding the exact tag command to be used: I don;t think it really matters (at least, as long as speed is concerned). Taskwarrior has to linearly parse the entire ~/.task/*.data files anyway to determine which tags are used; counting them while parsing should hardly give any overhead.

Ah good point.

In the end, I think it is good that you used "task tags" instead of "task _tags" (I think I actually recommended the opposite) because "task tags" returns only tags in the pending list, so it is faster. The consequence is, of course, if the VIT user is viewing a completed task and wants to remove a tag that is not present in their pending task list, it will not expand for them. But I don't think VIT even works well on completed tasks anyway, so I'm not worried about this.

$ time task tags

gives:

real	0m0.159s
user	0m0.140s
sys	0m0.020s


and

$ time task _tags

gives:

real	0m0.211s
user	0m0.200s
sys	0m0.008s

scottkosty:
Bas, this works great! I've tested for a while now and I find it to be very robust. Well done!

I think this will be a highlight when we announce the new features in VIT 1.3.

The following feature request falls in the category of "not that important so feel free to ignore but I'll mention it anyway":

If a user enters two tags, there is no tab-expansion for the second.

One issue I came across is that this slows down opening VIT for me. The following timings are actually an average of 5 runs.

Before the patch, when running "time vit<enter>Z" I get the following:

real	0m1.794s
user	0m1.692s
sys	0m0.104s

after the patch, I get the following:

real	0m1.978s
user	0m1.853s
sys	0m0.103s

Looking at the timings in my comment above, I would guess that the extra time is due to the call to "task tags", so I don't think there's anything we can do about it.

I have a few smaller task databases (e.g. that I use just for work), and I do not notice any slowdown (I did not time it though). So at least the extra time is a function of the database size and is not a flat penalty (which makes sense, but wanted to check to be sure).

So let's merge this.

What do you think of the following patch? It allows splitting on a combination of spaces and commas. I think some users may attempt to enter a comma-separated list of tags. The disadvantage of this patch is that if a user actually intended a comma as part of the tag, they will be surprised.

```
commit 67dfc0100cbda07f89d43e28b40618de165779dd
Author: Scott Kostyshak <scott.kostyshak@gmail.com>
Date:   Sat Jun 24 20:21:17 2017 -0400

    Allow tag list to be separated by commas
    
    The user might want to enter
    
        a, b, c
    
    or
    
        a,b,c
    
    and intend for the tags to be +a +b +c.
    
    This form of input may be intuitive to the user as it is what is
    displayed when running
    
        task <ID> edit

diff --git a/cmds.pl b/cmds.pl
index 66d7e83..0d08218 100644
--- a/cmds.pl
+++ b/cmds.pl
@@ -287,11 +287,12 @@ sub task_set_tag {
     return;
   }
 
-  # multiple tags can be specified separated by spaces
+  # multiple tags can be input separated by a combination of spaces and commas
+  #
   # keep track of the current modifier (default:+) and use it for subsequent tags
   # so "+a b c" means "+a +b +c" and "+a -b c" means "+a -b -c"
   my $mod='+';
-  foreach my $t (split(/\s+/,$tags)) {
+  foreach my $t (split(/,?\s+|,/,$tags)) {
     next if $t =~ m/^\s*$/;
 
     # check if a + or - was specified

```

Thanks also for your extra clean-up and for all of the comments. That really made it easy to read the code. It is nice to replace the hard-coded 410 with KEY_RESIZE. I think there are other places this is used. I will make the change for those places also.

If you're motivated to make a patch with getchar(), we could put that in at the beginning of the 1.4 cycle. But I want to keep the current code for 1.3, since it has been well-tested.  Note that @Ri-Ga had some comments on getchar() which you can see at 4de7351c (thanks for that great commit message, Richard!).

scottkosty:
> It is nice to replace the hard-coded 410 with KEY_RESIZE. I think there are other places this is used. I will make the change for those places also.

Done at c0614ef8.

scottkosty:
@baszoetekouw Can you take a look at the patch I suggest in https://github.com/scottkosty/vit/pull/5#issuecomment-310879911?

scottkosty:
@baszoetekouw any thoughts on the patch in https://github.com/scottkosty/vit/pull/5#issuecomment-310879911?

scottkosty:
I put the patch in at 2d6ec2cd.

PR #6, opened 2017-02-12
Add start/stop commands to Vit.

dbalan:
- bound to key 'b', starts or stops a task depending on its current status

More discussion here: https://bug.tasktools.org/browse/VT-131

### DIFF
--- a/cmds.pl
+++ b/cmds.pl
@@ -99,6 +99,36 @@ sub task_den_or_del {
 
 #------------------------------------------------------------------
 
+sub task_start_stop {
+    my ($ch, $str, $yes);
+    my $id = $report2taskid[$task_selected_idx];
+
+    my ($es, $result) = &task_exec("$id active");
+    my $prompt = "stop";
+    if ($es != 0) {
+        $prompt = "start";
+    }
+
+    $yes = &prompt_y("$prompt task $id?");
+
+    if (! $yes ) {
+        &draw_prompt_line('');
+        return;
+    }
+
+    my ($es, $result) = &task_exec("$id $prompt");
+    if ( $es != 0 ) {
+        $error_msg = $result;
+        &draw_error_msg();
+        return;
+    }
+    $feedback_msg = "Marked task $prompt ed.";
+    &draw_feedback_msg();
+    $reread_needed = 1;
+}
+
+#------------------------------------------------------------------
+
 sub task_done {
   my ($ch, $str, $yes);
   my $id = $report2taskid[$task_selected_idx];
--- a/cmds.pl
+++ b/cmds.pl
@@ -103,13 +103,16 @@ sub task_start_stop {
     my ($ch, $str, $yes);
     my $id = $report2taskid[$task_selected_idx];
 
-    my ($es, $result) = &task_exec("$id active");
+    my ($state, $_) = &task_exec("$id active");
     my $prompt = "stop";
-    if ($es != 0) {
+    $feedback_msg = "Stopped Task";
+
+    if ($state != 0) {
         $prompt = "start";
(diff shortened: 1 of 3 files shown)


dbalan:
@scottkosty Can you please merge this?

scottkosty:
I would first like some discussion on whether users prefer a toggle (like you propose) or two separate keybindings. I'd also like some feedback on the best keybindings. Once we implement a keybinding, it is not good to change it (because many will already be used to it). So it's best to make sure we get it right the first time.

I will start an email thread with VIT users to get feedback regarding a few issues. In particular, this one and #2.

dbalan:
Fair enough! Thanks for the clarification.

Quoting Scott Kostyshak (2017-02-25 00:10:25)

scottkosty:
I've finally taken a detailed look at your patch. It looks good! I especially
appreciate that you made the change to the 'commands' file, you used the
built-in prompt_y subroutine, and the style of your patch fits well with the
current style.

Thanks for your work on this useful feature!

I have the following minor comments. If you address the first 2 comments, I
will address comments 3 through 5 in follow-up commits. Or if you prefer, feel
free to do the others as well.

Comments:

1. Perl gives some warnings. They are always given but VIT clears the screen, so the easiest way to see them is:

$ vit -v
"my" variable $es masks earlier declaration in same scope at /usr/local/bin/vit line 390.
"my" variable $result masks earlier declaration in same scope at /usr/local/bin/vit line 390.
vit-1.3.dev (0fadb93)
$ 

The reason for these warnings is that you have the following lines:
    my ($es, $result) = &task_exec("$id active");
    my ($es, $result) = &task_exec("$id $prompt");

2. The second line of the git message should always be empty

3. add a changelog entry

4. The feedback after starting a task is currently
"Marked task start ed."

I would prefer for it to be

"Started task" and "Stopped task"

This is consistent with

  "Annotated..."

and the feedback from modify is

"Modified..."

This will involve a little annoying work since we cannot just append "ed" to
"stop". We must instead add an extra "p".

5. I would prefer to remove the task ID from the current prompt "start task
<TASKID>" so that it is only "start task?". The reason is partly to be
consistent (e.g. with 'm', 'p', and 'P'), and also because ideally I'd like to
abstract away from task ID numbers. They are just temporary numbers useful for
identifying a task, but in VIT we can identify the task with the cursor.
Ideally, it would be nice someday for the user to have the option of never
seeing a task ID while using VIT. Of course, sometimes it is nice to see a task
ID because the user might want to manually run a taskwarrior command on it.
Thus, we might allow the user to configure this behavior, and if so, the
behavior should be centralized (e.g. in prompt_y).

dbalan:
@scottkosty Thanks for reviewing!

I've made neccessary changes for this to work, please have a look?

scottkosty:
@dbalan I merged your branch and made some follow-up commits. They are all simply, but could you please take a look at the follow-up commits and confirm that they look OK to you?

Also, in the future, please format your commit messages according to common guidelines. See, for example: https://chris.beams.io/posts/git-commit/
In particular (taken from the URL):

1. Separate subject from body with a blank line
2. Limit the subject line to 50 characters
3. Capitalize the subject line
4. Do not end the subject line with a period
5. Use the imperative mood in the subject line
6. Wrap the body at 72 characters
7. Use the body to explain what and why vs. how

Also note: there's still a chance the keybinding will change. I'm planning to ask for feedback (either pre or post 1.3beta) before the final 1.3 release.

I'm sorry for my picky comments. I hope I don't scare you away from future contributions. Thank you f or this nice feature!

dbalan:
@scottkosty Apprciate the attention to detail. Also, thanks for the project!

PR #9, opened 2017-05-24
Add key 'l' for logging a task

Ri-Ga:
This makes "task log" as easily accessible as "task add". Would be nice if this feature could be included in the next version.

### DIFF
--- a/cmds.pl
+++ b/cmds.pl
@@ -16,7 +16,25 @@ sub prompt_quit {
 #------------------------------------------------------------------
 
 sub task_add {
-  my $str = &prompt_str("Add: ");
+  &task_cmd_parseable_with_prompt("add", "Add")
+}
+
+#------------------------------------------------------------------
+
+sub task_log {
+  &task_cmd_parseable_with_prompt("log", "Log")
+}
+
+#------------------------------------------------------------------
+
+# General function for invoking task commands with a prompt and
+# where the input provided at the prompt shall be parseable, i.e.,
+# be a sequence of strings. Examples are the commands "add" and
+# "log", while non-examples are "annotate" or "done".
+sub task_cmd_parseable_with_prompt {
+  my $cmd = $_[0];
+  my $prompt = $_[1];
+  my $str = &prompt_str("$prompt: ");
   if ( $str eq '' ) {
     &draw_prompt_line('');
     return;
@@ -24,7 +42,7 @@ sub task_add {
   # TODO: get rid off escaping by replacing the shell call in task_exec() by something like IPC::Open2().
   #       That would allow us to get rid of all shell expansions, quotations and escaping. [BaZo]
   $str =~ s{[&^\\]}{\\\&}g;
-  my ($es,$result) = &task_exec("add $str");
+  my ($es,$result) = &task_exec("$cmd $str");
   if ( $es != 0 ) {
     $error_msg = $result;
     &draw_error_msg();
--- a/commands
+++ b/commands
@@ -2,6 +2,7 @@
  cmd   a STRING        add a new task
  cmd   A STRING        add an annotation to the current task
  cmd   b STRING        start/stop a task
+ cmd   l STRINT        log a new task
  cmd   m STRING        modify the current task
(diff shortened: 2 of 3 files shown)


scottkosty:
@Ri-Ga Thanks for this well-written patch. I appreciate the comment explaining `task_cmd_parseable_with_prompt` and your remembering to patch the help file. I'm hesitant to use the keybinding 'l'. Right now, 'h' and 'l' are free, and we might want to bind them to some Vim-like behavior. I also wonder if we will use them for #2 (for example 'j' and 'k' could move among tasks and 'h' and 'l' could navigate among annotations within a task). We could use a different keybinding, such as 'L', but that is not very intuitive.

I see the point you make that 'task log' and 'task add' are similar commands so we should have support for both. But using 'task add' is much more common than 'task log'. And there is an easy way to add a keybinding in your .vitrc file:

````
map l=:! task log 
````
Using the .vitrc keybinding above is not as smooth as your patch because you do not get any feedback.

Right now I'm thinking to not include this in 1.3beta (and thus not in 1.3), but to add it on to the list of items to discuss in the near future. I'm planning on setting up an email group with VIT users and polling them for feedback. If we do that and this patch gets enough support, then we will try to find a way to incorporate it (i.e. perhaps not with 'l', although that is open to discussion as well). I know you wanted it in the next version, but I'm hesitant to assign some keybindings because once we assign them, it is difficult to change them to something else if we have a new feature that could use them, because many users will be surprised (and some .vitrc keybindings could be broken).

Ri-Ga:
I understand your concern.

What about adding the possibility to use something like `task_cmd_parseable_with_prompt` in .vitrc keybindings? This would allow users to decide for themselves but in a smoother way than the keybinding you suggest. One could then put my suggested mapping (binding the 'l' key to logging) into vitrc.5 as an example. What about something like

    map l=:pcmd log Log

So far, I have not yet had a look at the vit code related to the vitrc handling so I don't yet know how hard that is to implement. If you like the suggestion, I would try to prepare a patch, though.

scottkosty:
That's an interesting idea. I'm open to that. Another possible general solution would be to allow the user to bind keys to unbound functions. This would be consistent with mutt (another source of inspiration for VIT) and I believe Vit. So the user could do something like:
````
map l=task_log()
````
I have no problem with the function part of the code you proposed (although I would want to test and take a deeper look); it's just the key bind that concerns me.

What do you think about this approach compared to your approach? Actually, they are not mutually exclusive---it might be useful to implement both.

Let's think carefully about the interface before you work on a patch, and let's look at what similar software does.

I'm going to remove the 1.3 tag since I'd like to get a beta out soon. I'm sorry for this since I know you wanted it in the next version, but I think we want to think carefully about this and maybe get some guidance from others as well before your work on a general solution.

Thanks for this discussion!

Ri-Ga:
> Another possible general solution would be to allow the user to bind keys to unbound functions. This would be consistent with mutt

What do you mean by "unbound function"? (I never used mutt, so sorry if it's a thing that's clear to mutt users)

> What do you think about this approach compared to your approach? Actually, they are not mutually exclusive---it might be useful to implement both.

Sounds like an even more flexible approach. From a code cleanliness perspective I cannot judge this. From a user's perspective, it sounds nice.

> I'm going to remove the 1.3 tag since I'd like to get a beta out soon. I'm sorry for this since I know you wanted it in the next version, but I think we want to think carefully about this and maybe get some guidance from others as well before your work on a general solution.

Absolutely no problem. I certainly don't want to delay a 1.3 release.

scottkosty:
> What do you mean by "unbound function"? (I never used mutt, so sorry if it's a thing that's clear to mutt users)

For example ([source](http://vim.wikia.com/wiki/Mapping_keys_in_Vim_-_Tutorial_(Part_1))), here is a Vim keybinding:

````
:map! <F3> <C-R>=strftime('%c')<CR>
````

There is no key bound to the function strftime(), but the function still exists for the purposes above so that a user can choose to bind a key to it. "keybinding" is a technical term for what is usually referred to as a "keyboard shortcut", so "unbound" means that there is no keyboard shortcut for a function.

thehunmonkgroup:

PR #11, opened 2017-06-06
fix tab completion in prompt

Ri-Ga:
* implements that cycling through tab completions
   happens only when tab is pressed multiple times
   without other keys in between
 * adds a bit of code documentation to related variables
 * removes some old, commented code block for handling
   backspace in prompt (there already is was uncommented
   code for doing this handling)
 * addresses bug VT-11

### DIFF
--- a/prompt.pl
+++ b/prompt.pl
@@ -70,10 +70,10 @@ sub prompt_str {
   my ($prompt) = @_;
   $cur_pos = length($prompt);
   my $str = '';
-  my $tab_cnt = 0;
+  my $tab_cnt = 0;# number of subsequent tab presses (for iterating through completions)
+  my $tab_match_str = '';# value of $str upon first press of tab
   my $history_idx = 0;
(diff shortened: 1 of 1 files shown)


baszoetekouw:
note that this is also addressed in #5

Ri-Ga:
Indeed. I did not test your commit, but concerning functionality I'd be fine if then only your pull request is chosen and mine is closed. Two aspects of my commit that I do prefer over yours are

1. the fact that the counter (`$tab_cnt`) is reset at a single, central location in the code rather than scattered over all the individual character handlers;
2. the existence of a bit of code documentation (both for the variables as well as for the counter-resetting block), which I think really does not harm vit.

Would you mind including this in your pull request?

baszoetekouw:
Agreed.  I'll merge this PR into #5

Ri-Ga:
Just had a quick look: Great, thanks!

Closing this PR in favor of #5 .

PR #12, opened 2017-06-11
add "nowait" option to avoid asking for enter after modifications

### LINKED ISSUE #7: [Feature suggestion] Add a .vitrc variable to supress the confirmation shown after modifying and adding tasks
First, thanks for maintaining VIT! It is the best interface to taskwarrior that I have come across so far, and it really facilitates working with my tasks!

Whenever I add or modify a task in VIT, the shell pops up a message with info about the task and I need to press Enter to return back to VIT. I tried setting `verbose=nothing` in my `~/taskrc`, which removes the task information in the message, but still shows an empty message with one line prompting me to press return to continue.

Is there a way around this so that I just have to press return once when creating the task and stay in VIT, instead of seeing the message screen and having to press return to go back into VIT?

---

**Update 20170308** I initially forgot to mention that I have remapped task addition as such `map a=:!rw task add project:Work `, to save myself typing in the project every time. The default task addition binding does not require confirmation.

baszoetekouw:
This fixes #7

### DIFF
--- a/exec.pl
+++ b/exec.pl
@@ -27,6 +27,12 @@ sub task_exec {
 
 #------------------------------------------------------------------
 
+sub exited_successfully {
+  my $status = shift || 0;
+  return 1  if  WIFEXITED($status) and WEXITSTATUS($status)==0;
+  return undef;
+}
+
 sub shell_exec {
   my ($cmd,$mode) = @_;
   endwin();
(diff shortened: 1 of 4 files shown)


scottkosty:
merge-squashed at 35c70936.

scottkosty:
@baszoetekouw what do you think if I rename this config variable to just 'wait' ? I think that 'nowait=no' is slightly confusing. Also, 'wait' would be consistent with the recently added 'confirmation' variable.

scottkosty:
@baszoetekouw I was testing with

    set nowait=yes

If I press return on a task (to see more detailed information), nothing happens because it returns immediately. I'm wondering if we should make an exception for this case.

baszoetekouw:
I've changed the PR to now use "wait" instead of "nowait" (whcih I agree might be confusing), and I've fixed the bug for showing the summary with `wait=no`.
I'll make a new PR for these.

scottkosty:
Thanks, looks good.

Issue #145, opened 2018-02-12
[VT-124] vit [args] should be persistent


taskwarrior:
_David Patrick on 2015-08-13T20:58:39Z says:_ 

 vit can be launched with arguments, works great. For example;

vit rc.context=work

will launch vit with the work context active.

but the thing is, any subsequent action discards the original arguments, so by changing reports, I will lose the context setting.

I think it would be better if any arguments used in launching vit, remained persistent for that vit session, or until Ctrl L is used to refresh the screen. That way, a vit session could remain focused on whatever arguments were used to launch it.

taskwarrior:
Migrated metadata:
```
Created: 2015-08-13T20:58:39Z
Modified: 2015-08-15T19:56:05Z
```

taskwarrior:
_Scott Kostyshak on 2015-08-13T23:35:15Z says:_ 

 Good point. We'll have to think this through. One concern that comes to mind is that VIT would have to parse the command and figure out which part of the command is the filter and which part should remain persistent. I'm not sure this is easy, as Taskwarrior commands come in many ways.

taskwarrior:
_David Patrick on 2015-08-14T01:00:08Z says:_ 

 I think that if invoked with args, ALL of it should be presistent, filter, configs the works. Then any subsequent filter elements, while the app is running, should get added to those, but the original set remains until quit or ^L reload.

taskwarrior:
_Scott Kostyshak on 2015-08-14T05:32:50Z says:_ 

 But what happens when you change the filter? The other elements should stay right? But that is the problem -- it is hard to figure out what the other elements are by parsing the taskwarrior command. One way would be to ask Taskwarrior by help, e.g. by using --debug which shows what the filter is parsed as I think. I don't know if this should be done though.

taskwarrior:
_David Patrick on 2015-08-14T23:22:17Z says:_ 

 My thinking is that launching vit with arguments could (should?) mean "constrain this vit session to these variables", and that defines a BASE_FILTER set, which is used for that session. Any filters applied during the session (including those that might override) are applied on top of the base set.

This would allow a given vit instance to have a specified function. 

We've had this conversation before, and I still think it would be useful.

taskwarrior:
_Scott Kostyshak on 2015-08-15T01:18:22Z says:_ 

 Let's add some structure. Suppose I start VIT as:

vit test rc.abbreviation.minimum=4

Then I press 'f'.
[...]

Issue #136, opened 2018-02-12
[VT-115] Context support in vit


taskwarrior:
_David Patrick on 2015-02-25T04:25:28Z says:_ 

 Now that task has a function to filter by context (as of unreleased version 2.4.2) vit should have built-in support for the feature. 

For starters. a key-mapping " map @=:!wr task context" (in .vitrc) goes a long way, but tab-completion using the (new) _context helper-command, would be good.

taskwarrior:
Migrated metadata:
```
Created: 2015-02-25T04:25:28Z
Modified: 2015-07-28T21:39:26Z
```

taskwarrior:
_David Patrick on 2015-07-28T21:38:14Z says:_ 

 Related to context support; it's easy to change context with vit, but there's now visual feedback to indicate which, if any, context filter is in use.

Something like adding context (if any) to the title bar, would do it.

lenalebt:
I installed the latest development snapshot and do see the visual feedback in the title bar about the context I am in currently. However, the filters applied to the context do not seem to be used at all?

I have three contexts, and the applied filters do filter out all the other tasks. My main contexts are "work" and "home", and I'd like to keep them separate. Context switching and showing using taskwarrior CLI works as expected, but not in vit.

Am I missing something?

thehunmonkgroup:
VIT 2.x pulls data directly from the TaskWarrior database via the export command, and I expected that command to respect the set context, which it does not.

This means the needed fix will require manually parsing the context filters, storing them, and silently applying them to all commands, which may not be a simple fix. I'll see what I can do.

thehunmonkgroup:
The fix turned out to be fairly easy: https://github.com/scottkosty/vit/commit/fff7c1dff35746faeb0e6b2c29bcb1a93682c257

@lenalebt I'm going to wait a bit longer before rolling the next beta, but if you'd like to test this fix you should be able to install the latest code by uninstalling the previous beta version, then installing from the git 2.x branch:

```
pip uninstall vit
pip install git+https://github.com/scottkosty/vit.git@2.x
```

Let me know if that corrects the issue for you.

thehunmonkgroup:
I decided to finish this issue off and add a built in action for switching context, mapped to the 'c' key by default.

This opens the command bar, allowing you enter a context directly, and includes tab completion for all contexts and the 'none' keyword. The command respects the global 'wait' config setting.

[...]

PR #157, opened 2018-10-03
add option for default tags

thomasrebele:
I wanted to use taskwarrior also to organize tasks related to my job. I decided to put work-related and private tasks in one taskwarrior "session". My idea: use tags for hiding / showing the relevant tasks in vit.

scenario:
1. execute `vit --default-tags '+job'`
2. vit only shows tasks with tag `job`
3. add tasks. The new tasks will have the `job` tag


thomasrebele:
related issue: #145

scottkosty:
@thomasrebele thanks for your work on this! I don't have time to look at it now, but I made a note to look at it when I have time (this might not be for a while though).

noahfrederick:
@thomasrebele Maybe I'm misunderstanding the feature, but don't [Contexts](https://taskwarrior.org/docs/context.html) already cover (most of) your use case? E.g., `task context job && vit`.

Using contexts has advantages over this approach:

- Not exclusive to a specific UI such as vit
- More flexible (you can filter by tags, project, or any other criteria)

That being said, I'm looking forward to [better context support](https://github.com/scottkosty/vit/issues/136) in vit.

I realize that contexts don't really cover item 3 above, which is something I would find useful as well, but I'd much rather see it added upstream rather than to vit, and to be generalized to any properties, not just tags.

thomasrebele:
Indeed, the context feature is almost what I want. However, as @noahfrederick mentioned, it does not support setting tags automatically on newly created tasks. It would be nice to add such a feature to taskwarrior directly.

thomasrebele:
Revisiting this issue again: The taskwarrior hooks API allows to modify a task before adding it. This allows to set the tags based on the context. I implemented a script that analyzes the context definition and adds the tags.

You can find the script [here](https://github.com/thomasrebele/dotfiles/blob/master/config/terminal/taskwarrior.install/.task.dir/hooks.dir/on-add-set-context.symlink). Put it in the hooks dir (normally it's ~/.task/hooks), and use the context feature mentioned by @noahfrederick.

thehunmonkgroup:
@thomasrebele, looks like your hook script addresses the issue, so no need to reopen this, correct?

Also, I took a quick look at your script, you might change the line that finds the context to something more like:

```sh
task _show | grep "context=" | sed "s/context=//"
```

```task _show``` is nice in that it gives you much more machine readable output.

Also, rewriting JSON in Bash? That's bold ;) I think I would have gone with Python.

thomasrebele:
Thanks for the tip with `task _show`. I've changed my script accordingly.

I made the script work with sh (probably also POSIX compliant). That should make it future-proof ;)

Issue #151, opened 2018-02-12
[VT-130] vit --watch option


taskwarrior:
_David Patrick on 2017-02-01T18:35:44Z says:_ 

 While vit is running, if something changes with the data, vit has now way of knowing, and doesn't immediately reflect the change. With the --watch option, if any of the source files change, the screen gets reloaded. This will make vit more useful in a multi-wondow or multi-application scenario.

taskwarrior:
Migrated metadata:
```
Created: 2017-02-01T18:35:44Z
Modified: 2017-04-24T13:57:36Z
```

taskwarrior:
_David Patrick on 2017-02-01T18:39:26Z says:_ 

 related (obviates?) to VT-93 and VT-47

taskwarrior:
_Eric Hymowitz on 2017-02-03T20:31:23Z says:_ 

 I'm not (by trade) a programmer.  I'm not sure I understand your work flow or (for that matter) git.  But I have a possible patch.  It requires tons of testing.  Right now, it's not an option, it's just "on".  It should eventually be an option which sets the "$alarmtime" variable.  But in the meantime, it's working for me, and maybe somebody else can test it and expand on it and improve it.

{code}
diff --git a/env.pl b/env.pl
index 10c5b60..f7deb66 100644
* -- a/env.pl
+++ b/env.pl
@@ -59,6 +59,10 @@ sub init_task_env {
       print "$_\r\n";
       exit(1);
     }
+    if ( $_ =~ /data\.location\s+(.*)/ ) {
+      $data_location = $1;
+      $data_location =~ s/~/$ENV{'HOME'}/;
+    }
   }
   close(IN);
   if ( defined $vit_header_color ) {
@@ -107,6 +111,18 @@ sub init_task_env {
   push((at)reports,'summary');
   (at)report_types = sort((at)reports);

+
+  # check for file updates
+  $SIG{'ALRM'} = sub {
+    my $then = (stat("$data_location/pending.data"))[9];
+    if ( time() - $then < $alarmtime + 2)
+    {
+       # &init_curses('refresh');
+       &read_report('refresh');
+       &draw_screen();
+    }
+    alarm $alarmtime;
+  };
 }

 return 1;
diff --git a/vit.pl b/vit.pl
index 197bb65..14f2981 100755
* -- a/vit.pl
+++ b/vit.pl
[...]

PR #158, opened 2018-11-15
Add SIGUSR1 handler that reloads taskwarrior data

tarruda:
Also add taskwarrior on_exit hook that will ensure any running vit instances are notified.

This is a simpler alternative to the patch provided by #151. The idea is that the on_exit.vit script will be manually copied by the user to the ~/.task/hooks directory, and whenever taskwarrior is updated any running vit instances will be refreshed.

### DIFF
--- a/getch.pl
+++ b/getch.pl
@@ -5,8 +5,6 @@ sub getch_loop {
   while (1) {
     my $ch = $report_win->getch();
     &audit("Received key: $ch");
-    $refresh_needed = 0;
-    $reread_needed = 0;
     $error_msg = '';
     $feedback_msg = '';
 
@@ -295,6 +293,8 @@ sub getch_loop {
     $prev_ch = $ch;
     if ( $reread_needed ) { &read_report('refresh'); }
     if ( $refresh_needed || $reread_needed ) { &draw_screen(); }
+    $refresh_needed = 0;
+    $reread_needed = 0;
 
   }
 }
--- a/on-exit.vit
+++ b/on-exit.vit
@@ -1,0 +1,12 @@
+#!/bin/sh -e
+
+modified=0
+while read modified_task
+do
+	modified=1
+done
+
+if [ $modified -eq 1 ]; then
+	# notify all running vit instances
+	killall -USR1 vit 2> /dev/null || true
+fi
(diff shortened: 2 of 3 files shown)


scottkosty:
Thanks for looking into this! I will take a look next time I process Vit issues. Probably the first thing I'll want to do is release 1.3 and then I'll be more open to features like this. I look forward to testing it. In the mean time, let me know if you make any modifications or find any unexpected behavior with your locally-applied patch.

tarruda:
Thanks

There are a few caveats with this approach:

- It depends the `getch` calls being interrupted when it receives SIGUSR1. I don't know if this behavior is portable across all operating systems or curses implementation (on Debian, the Perl curses module is linked against libncursesw and it works fine).
- `getch` will return -1 when interrupted by the signal, but this is currently is only handled in the getch_loop function. That means when a signal is received and vit is displaying a prompt or is in command mode, a `-1` will be echoed and the screen won't be updated.
- Unlike with the patch from #151, changes to .taskrc are not reflected until a task is updated.

There might be more issues I'm not aware, but this patch works fine for me.

The ideal way of implementing async updates from other programs is to use a function like `select()` to watch multiple file descriptors, but I'm not experienced with Perl so my goal was to implement the functionality with the minimum changes.

scottkosty:
Thanks for all of those comments! That will be helpful to me when reviewing the code and figuring out what to do. I might ask you to add an option to vitrc that turns this behavior on. i.e., it would be off by default but could be turned on with a simple line in the vitrc. I'm not sure though so don't spend your time on that yet.

ahillio:
Is this PR related in functionality to https://github.com/scottkosty/vit/pull/191 ?  I'd love to see `vit` screens automatically refreshed when the underlying task data is changed (which happens for me frequently using vimwiki diary (w/taskwiki plugin) in one tmux pane and vit in another tmux pane).

thehunmonkgroup:
This PR was for the no longer supported 1.3.x branch. I think #191 would effectively replace it, but not completely sure.

thehunmonkgroup:
Can be accomplished with the auto-refresh tools: https://github.com/vit-project/vit/blob/2.x/CUSTOMIZE.md#auto-refreshing-vits-interface

PR #159, opened 2018-11-17
Allow custom .vitrc location with VITRC env var.

tarruda:


scottkosty:
Makes sense, and is consistent with other programs (e.g., `TASKRC` environment variable for Taskwarrior). I also like the change from `~` to using the `HOME` environment variable.

I plan to merge this sometime after releasing 1.3. Thanks!

thehunmonkgroup:

PR #164, opened 2019-01-13
Only run task sync if a taskd server is configured

jspricke:
Allows to reread the tasks using with no taskd server using 's'.


thehunmonkgroup:

Issue #19, opened 2017-07-14
backtick breaking insertion


mellotanica:
automated insertions (e.g. command `a`, `A`) are vulnerable to backtick ` character, if one writes a line of text containing two backticks, the program tries to execute what stands between them, like bash would do.

maybe strings should be sanitized in some way before being passed to task when executing these commands?

obviously this behavior is correct when using `:!` or `t` commands

scottkosty:
> automated insertions (e.g. command a, A) are vulnerable to backtick ` character

For 'A' this should not be the case. Are you sure you're using the 1.3 branch?

For 'a', the backticks are indeed interpreted by the shell, because we want to allow Taskwarrior to parse the command (e.g. we want Taskwarrior to interpret "due:3days" for the add command, as opposed to the annotate command). Instead of sanitizing the strings, which I think could be complicated to do perfectly, we could have two prompt modes, one where everything is quoted (which is the current case of annotate), and one where everything is passed to the shell (which is the current case of add), and the user would be able to easily switch between the modes.

mellotanica:
> For 'A' this should not be the case. Are you sure you're using the 1.3 branch?

it actually does not work properly, i'm on 1.3, if i try to annotate a task with something like
```annotation with `backticks` works?```
the task results annotated with
```annotation with  works?```
it does not complain about `batkticks` not being a known executable like 'a' does, but it still tries to execute the command, in fact leaving an annotation like
```annotation with `pwd` works?```
results in 
```annotation with /data/repos/vit works?```


being able to switch between the two versions would be nice, for 'a' in "sanitized" mode one could also prompt for a second line of arguments after the task name, allowing the additional parameters to be passed to task

scottkosty:
hmmm OK then with A something is broken. For me, if I do A and then

    annotation with `pwd` works?

I get that as the literal annotation, no pwd command is run. I'd like to figure out why we see differences. What is the output for you of the following command?

    vit --version



> one could also prompt for a second line of arguments after the task name, allowing the additional parameters to be passed to task

This is an interesting idea. I didn't think about that. So you are suggesting a double prompt. In the first prompt the user would enter the task description, and for this prompt the default would be to have everything escaped. For the second prompt, the user would enter a string that is meant to be parsed by Taskwarrior (e.g. "due:3days wait:1day"). The second prompt would not be quoted when passed. Did I understand that correctly?

scottkosty:
Note that part of the discussion here is related to #15.

[...]

PR #165, opened 2019-02-18
Fix empty tasklist

thomasrebele:
Starting vit without any tasks leads to an annoying error message:

    Fatal error: default.command ("next") must print an "ID" column

The first commit ignores that the columns are missing if the task command outputs `No matches`.
The second commit adds an option to update the task list, even if there are no tasks in the report.

### DIFF
--- a/env.pl
+++ b/env.pl
@@ -20,6 +20,7 @@ sub init_shell_env {
 sub init_task_env {
   my @reports;
   my $id_column = 0;
+  my $empty_report = 0;
   my ($header_color,$task_header_color,$vit_header_color);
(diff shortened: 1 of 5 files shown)


thomasrebele:
I just found an issue that discusses about escaping: #19 
I guess it would be nice to make the escaping of the different commands consistent.

scottkosty:
@thomasrebele thanks for your patches. I've finally released VIT 1.3. I hope to find time to review your patches in the coming weeks. Thank you for your time on these!

thehunmonkgroup:

PR #166, opened 2019-02-18
improve escaping of special characters when adding tasks

thomasrebele:
This commit improves escaping when adding a task.
Most notably, adding task descriptions with parenthesis is now supported.


thehunmonkgroup:

PR #167, opened 2019-03-03
remove hard-coded perl path

thehunmonkgroup:
This removes the hard-coded /usr/bin/perl from the shebang
line, in favor of a /usr/bin/env approach, which allows for
non-standard Perl locations, such as /usr/local/bin/perl
(a requirement for building on OS X).

Since /usr/bin/env argument parsing isn't portable, the -w
argument in the shebang line is removed in favor of
'use warnings', which is the recommended approach given at
https://perldoc.perl.org/warnings.html

Tested as working on Ubuntu and OS X

### DIFF
--- a/configure
+++ b/configure
@@ -1771,11 +1771,6 @@ fi
 if test "$PERL" = "NOT_FOUND"; then
   as_fn_error $? "can not continue: perl not found" "$LINENO" 5
 fi
-if test "$PERL" != "/usr/bin/perl"; then
-  echo configure: error: can not continue... perl must be installed as /usr/bin/perl
-  echo confgiure: hint: try \"sudo ln -s $PERL /usr/bin/perl\"
-  exit 1
-fi
 
 PERL_MODULE_CHECK=`perl -e "use Curses" 2> /dev/null; echo $?`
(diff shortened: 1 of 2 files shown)


scottkosty:
@thehunmonkgroup thanks for the patch and for your clear description. I've used the `/usr/bin/env` approach in other scripts with success for years. I've only recently become aware of some potential downsides (noted, e.g., [here](https://unix.stackexchange.com/questions/29608/why-is-it-better-to-use-usr-bin-env-name-instead-of-path-to-name-as-my)). I still think your patch leads to an improvement for most users, so I am leaning towards including it. I'll think some more and then probably merge it. Thanks for your work.

thehunmonkgroup:
That thread is interesting, seems like it's largely a matter of opinion, since there are advantages and disadvantages to both approaches.

There seem like two other reasonable options to increase portability:

1. This approach outlined in the thread:
  ```
  #!/bin/sh
  exec perl -x "$0" "$@"
  #!perl
  ```
2. Handling setting the shebang line as part of the configure/make process.

thehunmonkgroup:

Issue #39, opened 2018-02-12
[VT-17] vit; % should act on uuids not ids


taskwarrior:
_David Patrick on 2013-09-08T17:23:44Z says:_ 

 Keybindings and commands can acts on the selected task, as represented in the .rc by "%=ids", but this is sub-optimal for 3 reasons;

* all tasks have uuids, but only pending tasks have ids
* ids can be changed externally (cronjob, cli, sync) without vit knowing about it.
* the % keybind token (and :! command) only works for reports that display ID

taskwarrior:
Migrated metadata:
```
Created: 2013-09-08T17:23:44Z
Modified: 2014-02-09T01:16:50Z
```

taskwarrior:
_Scott Kostyshak on 2013-12-12T23:43:10Z says:_ 

 I agree this should be implemented, but unless I receive a patch this will have to wait.

thehunmonkgroup:
VIT 2.x alpha 1 has been released, and uses uuids for all task interactions

https://github.com/scottkosty/vit/releases/tag/v2.0.0a1

PR #173, opened 2019-05-12
Use UUID for task commands (Closes: #39)

jspricke:


scottkosty:
@jspricke thanks for this pull request! I think we should consider this for 1.4.

thehunmonkgroup:

PR #188, opened 2019-07-26
AUTHORS.md: clarify roles

scottkosty:
@thehunmonkgroup 
Chad, what do you think of these changes? Additionally, it seems strange for me to have the role of "Maintainer" since you are currently doing more maintaining that I am. Either we should add "Maintainer" to your role or subtract it from mine. I don't have a preference which.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -1,8 +1,8 @@
 The development of VIT was made possible by the significant
 contributions of the following people:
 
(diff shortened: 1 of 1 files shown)


thehunmonkgroup:
I'd say add Maintainer to my list of roles in addition to the changes you have so far, and commit it to the 2.x branch. Sound good?

scottkosty:
Sounds good.

PR #162, opened 2019-01-12
Refresh every 30 minutes

jspricke:

Issue #69, opened 2018-02-12
[VT-47] vit; reread command and reread.freq=60 .vitrc setting


taskwarrior:
_David Patrick on 2013-09-08T01:05:26Z says:_ 

 Vit has now idea if something else changes the task data. It's even possible for vit to make a change it doesn't know about 
(for this reason, some keybinds should use the --reread option) 
and I think some reread/refresh/reload command would be a good idea.

A better idea is a .vitrc auto-reread setting as number of seconds between auto-reread.

   auto-reread=60

might be a good default.

taskwarrior:
Migrated metadata:
```
Created: 2013-09-08T01:05:26Z
Modified: 2017-02-01T19:08:19Z
```

taskwarrior:
_Scott Kostyshak on 2015-02-28T23:20:55Z says:_ 

 There is a way (inotify) for VIT to know whenever one of the .data files changes. If that is implemented, would you still want to implement reread.freq ?

You also bring up another question. If VIT uses inotify as described above, should --reread be deprecated?

taskwarrior:
_Jochen Sprickerhof on 2015-12-01T09:34:33Z says:_ 

 I've written a simple patch to update every 30min. Having an option to configure the timeout would be great, but would need some more coding.
I'm not sure about using inotify, as it would not work if you use taskd to sync.

taskwarrior:
_Scott Kostyshak on 2015-12-02T03:11:35Z says:_ 

 Why wouldn't the inotify implementation work if taskd is used to sync? Wouldn't a taskd sync change the .data file and thus inotify would be triggered?

pbeckingham:
[VT-47_0001-Refresh-and-sync-every-30-minutes.patch.txt](https://github.com/scottkosty/vit/files/1755286/VT-47_0001-Refresh-and-sync-every-30-minutes.patch.txt)

thehunmonkgroup:
Can be accomplished with the auto-refresh tools: https://github.com/vit-project/vit/blob/2.x/CUSTOMIZE.md#auto-refreshing-vits-interface

PR #189, opened 2019-07-27
Add option to reload the tasks every n minutes

jspricke:
Reimplementation of #162 for 2.x

### DIFF
--- a/vit/application.py
+++ b/vit/application.py
@@ -134,6 +134,7 @@ class Application():
     def setup_config(self):
         self.confirm = self.config.confirmation_enabled
         self.wait = self.config.wait_enabled
+        self.update = self.config.update
 
     def register_managed_actions(self):
         # Global.
@@ -392,7 +393,7 @@ class Application():
                     metadata.pop('uuid')
             elif command in self.reports:
                 self.extra_filters = args
-                self.update_report(command)
+                self.update_report(report=command)
                 if 'uuid' in metadata:
                     metadata.pop('uuid')
             else:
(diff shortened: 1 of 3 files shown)


scottkosty:
Thanks for porting your patch to VIT 2.x! I still want to understand a couple of things: First, if VIT auto-updated every time one of the Taskwarrior *.data files changed, would this be sufficient or would you still want to set a custom timer? We have some discussion on this at #69, and you brought up a point about that not working if you use taskd to sync but I don't understand why that's the case---if taskd syncs, doesn't that update the *.data files?

jspricke:
file changes: VIT uses URWID for the main loop, where you can add alarms and watches: http://urwid.org/reference/main_loop.html, but the `watch_file()` only takes a `fd`, and I didn't got it to work with the `*.data` files, so I went with the alarm. I think it would make more sense to implement a lazy reload in the tasklib (I did a time based check in my task lib here: https://github.com/jspricke/python-icstask/blob/master/icstask.py#L54).

taskd: If you have an other system pushing changes to taskd, VIT needs to sync those to the local `~/.task` to be able to display them. But I gave up on taskd in the mean time, this is why I didn't add the sync in this implementation.

scottkosty:
@jspricke I see. But suppose (just theoretically) that VIT could detect any change to the *.data files, and whenever a change was detected, VIT refreshes. Would that solve the problem?

jspricke:
> Would that solve the problem?

Yes, implemented this in #191 now. Feel free to merge #191 and close 
this.

scottkosty:
> Yes, implemented this in #191 now.

Cool! Nice work. I have comment, which I'll put over there.

thehunmonkgroup:
Unless there's a use case where a time-based approach has advantages over an event-based approach, which I'm not seeing, #191 should be the way we go.

PR #191, opened 2019-07-28
Reload events if pending.data changed

jspricke:
This needs the Python watchdog library.

As discussed in #69, alternative to #189.

### DIFF
--- a/vit/config/config.sample.ini
+++ b/vit/config/config.sample.ini
@@ -57,6 +57,14 @@
 #included in the active filter for this setting to have effect.
 #focus_on_add = False
 
+# Boolean. If true, VIT will monitor TaskWarrior's pending.data file, and
+# reload the currently displayed report if the file is changed. For most
+#operating systems this is done by a kernel API and does not need extra
+#resources. See
+#https://pythonhosted.org/watchdog/installation.html#supported-platforms-and-caveats
+#for more information.
+#auto_reload = True
+
 [report]
 
 # The default Taskwarrior report to load when VIT first starts, if no report
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
 tasklib>=2.4.3
 urwid>=2.1.2
 backports.zoneinfo;python_version<"3.9"
+watchdog
--- a/vit/application.py
+++ b/vit/application.py
@@ -5,6 +5,7 @@ from importlib import import_module
 import subprocess
 # TODO: Use regex module for better PCRE support?
 #       https://bitbucket.org/mrabarnett/mrab-regex
+import os
 import re
 import time
 import copy
@@ -12,6 +13,8 @@ from inspect import isfunction
 from functools import reduce
 
 import urwid
+from watchdog.observers import Observer
+from watchdog.events import PatternMatchingEventHandler
 
 from vit import version
 from vit.exception import VitException
@@ -341,6 +344,7 @@ class Application():
                 # individual edit actions consistent, recurrence.confirmation
                 # is set to 'no', so that only the edited recurring task is
                 # modified.
+                self.model.skip_watchdog = True
                 returncode, stdout, stderr = self.command.run(['task', 'rc.recurrence.confirmation=no', metadata['uuid'], 'modify', 'wait:%s' % wait], capture_output=True)
                 if returncode == 0:
                     self.table.flash_focus()
@@ -576,6 +580,24 @@ class Application():
     def init_task_list(self):
         self.model = TaskListModel(self.task_config, self.reports)
 
+    def observer_update(self, _):
+        if not self.model.skip_watchdog:
+            self.update_report()
+        self.model.skip_watchdog = False
+
+    def init_observer(self):
+        pipe = self.loop.watch_pipe(self.observer_update)
+
+        def on_modified(event):
+            os.write(pipe, b'x')
+
+        event_handler = PatternMatchingEventHandler(patterns=['*/pending.data'], ignore_directories=True)
+        event_handler.on_modified = on_modified
+
+        self.observer = Observer()
+        self.observer.schedule(event_handler, self.model.data_location)
+        self.observer.start()
+
     def init_autocomplete(self):
         context_list = list(self.contexts.keys()) + ['none']
         self.autocomplete = AutoComplete(self.config, extra_filters={'report': self.reports.keys(), 'help': self.help.autocomplete_entries(), 'context': context_list})
@@ -626,6 +648,7 @@ class Application():
 
     def execute_command(self, args, **kwargs):
         update_report = True
+        self.model.skip_watchdog = True
         wait = True
         if 'update_report' in kwargs:
             update_report = kwargs.pop('update_report')
@@ -957,6 +980,8 @@ class Application():
         if report:
             self.report = report
         self.init_task_list()
+        if self.config.get('vit', 'auto_reload'):
+            self.init_observer()
         self.build_frame()
         self.widget = MainFrame(
             urwid.ListBox([]),
--- a/vit/config_parser.py
+++ b/vit/config_parser.py
@@ -46,6 +46,7 @@ DEFAULTS = {
         'mouse': False,
         'abort_backspace': False,
         'focus_on_add': False,
+        'auto_reload': True,
     },
     'report': {
         'default_report': 'next',
--- a/vit/task.py
+++ b/vit/task.py
@@ -18,6 +18,7 @@ class TaskListModel(object):
         self.reports = reports
         self.report = report
         self.tasks = []
+        self.skip_watchdog = False
         if report:
             self.update_report(report)
 
@@ -69,6 +70,7 @@ class TaskListModel(object):
     def task_description(self, uuid, description):
         task = self.get_task(uuid)
         if task:
+            self.skip_watchdog = True
             task['description'] = description
             task.save()
             return task
@@ -77,6 +79,7 @@ class TaskListModel(object):
     def task_annotate(self, uuid, description):
         task = self.get_task(uuid)
         if task:
+            self.skip_watchdog = True
             task.add_annotation(description)
             return task
         return False
@@ -84,6 +87,7 @@ class TaskListModel(object):
     def task_denotate(self, uuid, annotation):
         task = self.get_task(uuid)
         if task:
+            self.skip_watchdog = True
             task.remove_annotation(annotation)
             return task
         return False
@@ -91,6 +95,7 @@ class TaskListModel(object):
     def task_priority(self, uuid, priority):
         task = self.get_task(uuid)
         if task:
+            self.skip_watchdog = True
             task['priority'] = priority
             task.save()
             return task
@@ -99,6 +104,7 @@ class TaskListModel(object):
     def task_project(self, uuid, project):
         task = self.get_task(uuid)
         if task:
+            self.skip_watchdog = True
             task['project'] = project
             task.save()
             return task
@@ -107,36 +113,43 @@ class TaskListModel(object):
     def task_done(self, uuid):
         task = self.get_task(uuid)
         if task:
+            self.skip_watchdog = True
             try:
                 task.done()
                 return True, task
             except (Task.CompletedTask, Task.DeletedTask) as e:
+                self.skip_watchdog = False
                 return False, e
         return False, None
 
     def task_delete(self, uuid):
         task = self.get_task(uuid)
         if task:
+            self.skip_watchdog = True
             try:
                 task.delete()
                 return True, task
             except Task.DeletedTask as e:
+                self.skip_watchdog = False
                 return False, e
         return False, None
 
     def task_start_stop(self, uuid):
         task = self.get_task(uuid)
         if task:
+            self.skip_watchdog = True
             try:
                 task.stop() if task.active else task.start()
                 return True, task
             except (Task.CompletedTask, Task.DeletedTask, Task.ActiveTask, Task.InactiveTask) as e:
+                self.skip_watchdog = False
                 return False, e
         return False, None
 
     def task_tags(self, uuid, tags):
         task = self.get_task(uuid)
         if task:
+            self.skip_watchdog = True
             for tag in tags:
                 marker = tag[0]
                 if marker in ('+', '-'):


scottkosty:
Nice work! On a quick read, one comment I have is that the patch only watches the file pending.data. I think this makes sense for the large majority of use cases, but I wonder if we might as well watch more. For example, the user on computer1 could be viewing the "completed" report in VIT, which I imagine reads from the file completed.data. Suppose that on computer2 you created and completed a task and then synced it. Then on computer1 you sync. In that case, I'm not sure but I would guess that the pending.data file would not change and the completed.data file would change.

The above is just a minor comment, but I figure we might as well discuss it now. Please don't put in the time to make the actual change. I'm not sure my comment is valid so we should discuss first. I can't think of a related comment regarding the undo.data or the backlog.data files (that is, I don't think we would want to reread in these cases).

As for merging the patch into 2.x, we would have to wait until either I felt comfortable with the new code base or if Chad had the time to take a look. I also don't know how big the disadvantage is to relying on a new dependency.

I do hope that we can eventually merge this because the feature has been requested by several different users and there are a few related bug/feature reports. Thanks a lot for your time and patience on this, Jochen!

jspricke:
>Nice work! On a quick read, one comment I have is that the patch only watches the file pending.data. I think this makes sense for the large majority of use cases, but I wonder if we might as well watch more. For example, the user on computer1 could be viewing the "completed" report in VIT, which I imagine reads from the file completed.data. Suppose that on computer2 you created and completed a task and then synced it. Then on computer1 you sync. In that case, I'm not sure but I would guess that the pending.data file would not change and the completed.data file would change.

I don't have a taskd running, could you (or someone else) maybe test 
this scenario?

We can watch both files, but would get an event for every changed file. 
This results in two events and update cycles, if you complete a task 
from the command line. I don't see an easy way to combine both events 
again, but maybe we can change the watched file depending on the viewed 
report (haven't looked into the code)?

thehunmonkgroup:
Thanks for taking the time to put this together. While I can imagine the use cases for this feature, I can also imagine that it might be a nightmare to support, and I'm not inclined to commit this to VIT core.

My long-term plan for VIT is to include support for plugins, but I haven't put in the time to work through the plugin architecture. This kind of feature seems ideally suited as a plugin.

I hadn't planned to implement the plugin stuff until VIT 3.x, but perhaps a simple initial implementation would be a reasonable first step. The downside is that plugins written for VIT 2.x would most likely not be compatible with VIT 3.x.

thehunmonkgroup:
This PR isn't going to make it into the 2.0.0 release, but I definitely want to get back to it in the future -- I think the VIT plugin architecture needs to make it further along first...

jspricke:
Do you have any proposal how to move this forward? I'm using (and rebasing ;) ) this for years without problems.
Would it be an option to make this optional and disabled by default?

thehunmonkgroup:
I still prefer my original suggestion of making it a plugin. However, I have not had time to begin working on a plugin architecture for VIT, nor will I in the foreseeable future.

I don't understand exactly how this code works, and don't really have the time now to invest in understanding it.

I'm also not clear how many people really need this feature.

All that said, I do understand it serves a useful purpose for some users, so here's my proposal, and if you are amenable to the terms, we can look at moving forward with this PR:

1. This will not go in until *after* a 2.2.0 release is cut, which I can hopefully make happen soon.
2. I do not want to maintain this particular piece of code, and so will expect others to provide fixes if/when it needs any
3. If number two above doesn't happen, I will most likely remove the feature instead of fix it myself
4. If/when VIT gets a proper plugin architecture, I would want this feature moved to a plugin.

jspricke:
>All that said, I do understand it serves a useful purpose for some users, so here's my proposal, and if you are amenable to the terms, we can look at moving forward with this PR:
>
>1. This will not go in until *after* a 2.2.0 release is cut, which I can hopefully make happen soon.
>2. I do not want to maintain this particular piece of code, and so will expect others to provide fixes if/when it needs any
>3. If number two above doesn't happen, I will most likely remove the feature instead of fix it myself
>4. If/when VIT gets a proper plugin architecture, I would want this feature moved to a plugin.

That would be fine with me.

thehunmonkgroup:
Splendid. Please ping me again on this PR after 2.2.0 gets released.

thehunmonkgroup:
`2.2.0` has been released. Before I commit this PR as promised, I'd like to explore one other dynamic: should this feature be disabled by default, and enabled via a config option?

I'm not sold on the idea of it being configurable, but I am wondering if this creates some overhead for a feature that most users won't need or care about.

jspricke:
>`2.2.0` has been released. Before I commit this PR as promised, I'd like to explore one other dynamic: should this feature be disabled by default, and enabled via a config option?
>
>I'm not sold on the idea of it being configurable, but I am wondering if this creates some overhead for a feature that most users won't need or care about.

That would be easy to do but let's discuss this first. Do you have any 
reasoning why users won't need or care about this?

I had a look through the bug reports and found a number of related 
issues:

https://github.com/vit-project/vit/issues/45
https://github.com/vit-project/vit/issues/69
https://github.com/vit-project/vit/issues/112
https://github.com/vit-project/vit/issues/115
https://github.com/vit-project/vit/issues/127
https://github.com/vit-project/vit/issues/151
https://github.com/vit-project/vit/issues/169
https://github.com/vit-project/vit/pull/158
https://github.com/vit-project/vit/pull/162

And taskwarrior-tui has this feature as well:

https://github.com/kdheepak/taskwarrior-tui

Also note that this does not interfere with user input, so for example 
when adding or viewing a task, the user input will stay.

And the python-watchdog library has support for all major operating 
systems with fallback options:

https://pythonhosted.org/watchdog/installation.html#supported-platforms-and-caveats

Finally, let me describe how I use vit and why the features is useful to 
me (and hopefully others):

I have vit constantly running in a terminal multiplexer (tmux) on my 
server so I can ssh in at any time to see the next tasks. At the same 
time I have a number of tools running to sync the task DB with other 
systems. I use bugwarrior to sync it to a number of issue tracks, 
icstask to sync it to my smartphone, some tasks are modified from the 
console or other scripts.. So for me it is important that vit is always 
up to date with what is in the task DB. 
I think for casual users it could even be confusing that vit could show 
outdated information and modifying tasked based on that could lead to 
problems.

thehunmonkgroup:
> That would be easy to do but let's discuss this first. Do you have any reasoning why users won't need or care about this?

The most basic reason is that this would be a fundamental change in behavior, and not everybody may like it. There are absolutely scenarios where a user would not want to be surprised to find what was on the screen a minute ago is not on the screen now.

I'm also not clear on the performance cost of running the watchdog all the time. I'm guessing it's minor, but then there are users that complain that the report load times are too slow, and some people don't want any unnecessary processes running, no matter how light.

All that being said, I'm guessing the performance hit is almost zero, and most people don't care if a new watchdog process is running, and they probably would prefer to see the most recent state of the task database on their screen, even if it means it being refreshed automatically.

So, my proposal is: add a configuration option to disable it, make the configuration ON by default, note the configuration option clearly in [config.sample.ini](https://github.com/vit-project/vit/blob/2.x/vit/config/config.sample.ini), and the implications of it being both enabled and disabled.

jspricke:
> I'm also not clear on the performance cost of running the watchdog all the time. I'm guessing it's minor, but then there are users that complain that the report load times are too slow, and some people don't want any unnecessary processes running, no matter how light.

For most operating systems the watchdog is provided by the kernel, so no extra costs.

> So, my proposal is: add a configuration option to disable it, make the configuration ON by default, note the configuration option clearly in [config.sample.ini](https://github.com/vit-project/vit/blob/2.x/vit/config/config.sample.ini), and the implications of it being both enabled and disabled.

I've added a `disable_watchdog` option.

scottkosty:
> The most basic reason is that this would be a fundamental change in behavior, and not everybody may like it. There are absolutely scenarios where a user would not want to be surprised to find what was on the screen a minute ago is not on the screen now.

+1. I'm not sure if I will want it enabled for my personal use. Part of the reason for me is that my Taskwarrior database is big enough, and my non-SSD HD slow enough, that it takes a few seconds to refresh my tasks (i.e., to reload VIT). It's really poor practice on my part, but I haven't ~~had~~ made time to clean things up in my Taskwarrior data base.

> So, my proposal is: add a configuration option to disable it, make the configuration ON by default, note the configuration option clearly in [config.sample.ini](https://github.com/vit-project/vit/blob/2.x/vit/config/config.sample.ini), and the implications of it being both enabled and disabled.

Makes sense. I'm not sure if I understood correctly, but just in case: Instead of having something like `disable-auto-reload=no` (i.e., a double-negative) I think it's generally easier to understand `enable-auto-reload=yes` and `enable-auto-reload=no`.

Another point in favor of the proposal by @thehunmonkgroup to disable the feature by default is that we can always flip that switch (i.e., change the default to enable it) in the future based on feedback. I imagine it will work smoothly, but it's often good practice to only flip that switch after considerable testing has been done, and I think only a couple of people have tested this feature so far.

By the way, I'm not a fan of the config name having "watchdog". I think we should separate interface from implementation. I proposed "enable-auto-reload" above. If both of you like "watchdog" in the config though, that's fine with me.

Thank you @jspricke and @thehunmonkgroup. This has been a request by many for many years. I think a lot of VIT users will appreciate this.

jspricke:
> +1. I'm not sure if I will want it enabled for my personal use. Part of the reason for me is that my Taskwarrior database is big enough, and my non-SSD HD slow enough, that it takes a few seconds to refresh my tasks (i.e., to reload VIT). It's really poor practice on my part, but I haven't ~had~ made time to clean things up in my Taskwarrior data base.

Note that you would only notice it if you change the database from the command line and then directly switch to a running vit.

Maybe check the size of your `completed.data` and rename it if it got too big ;).

> Makes sense. I'm not sure if I understood correctly, but just in case: Instead of having something like `disable-auto-reload=no` (i.e., a double-negative) I think it's generally easier to understand `enable-auto-reload=yes` and `enable-auto-reload=no`.

Good point, renamed to `auto_reload`.

vit/config/config.sample.ini
```diff
@@ -57,6 +57,9 @@
 #included in the active filter for this setting to have effect.
 #focus_on_add = False
 
+# Boolean. If true, VIT will reload if the tasks where changed on disk.
```
thehunmonkgroup: I think it might be beneficial to be a bit more descriptive here, since this is the only documentation we're adding at the moment.

First proposal:

Boolean. If true, VIT will monitor TaskWarrior's pending.data file, and reload the currently displayed report if the file is changed.

jspricke: Added, plus some extra information. Thanks!

thehunmonkgroup:
I can't believe I neglected to ask this before, but how does this not end up loading a report twice? 
1. Person edits task in VIT
2. task database is modified
3. VIT reloads report
4. watcher notices, reloads report.

jspricke:
Good point, I've adopted `update_report()` to only reload once the data changed on disk.

scottkosty:
> I can't believe I neglected to ask this before, but how does this not end up loading a report twice?

Nice catch. I didn't think of that either :)

scottkosty:
> Note that you would only notice it if you change the database from the command line and then directly switch to a running vit.

Indeed I find myself doing this frequently. I should figure out why in some cases I prefer command line instead of VIT, and should try to implement whatever feature would encourage me to stay in the world of VIT. I really don't know why I do that.

> Maybe check the size of your `completed.data` and rename it if it got too big ;).

Good idea, I should do that.

thehunmonkgroup:
What's with the `pipe` variable in `update_report`, and how it's used in `build_main_widget`?

I've read that a couple times and still don't get it.

jspricke:
> What's with the `pipe` variable in `update_report`

That is passed from the urwid.MainLoop.watch_pipe callback:

https://github.com/urwid/urwid/blob/master/urwid/main_loop.py#L211

> and how it's used in `build_main_widget`?

With `auto_reload=True`, `update_report` is skipped when called without the pipe parameter and executed through the watchtdog instead (with `pipe` set). Upon initial load in `build_main_widget`, `pipe` is set to not skip `update_report`.

thehunmonkgroup:
I'm actually not crazy about the way the `update_report()` signature is altered. Isn't there a way we can do this and preserve report as the first arg? My preference would be to not change that function at all, but if that's not possible, then at least let's preserve existing argument order.

thehunmonkgroup:
And in `on_modified` you're just writing garbage to the pipe, because it's not needed?

thehunmonkgroup:
Another concern: your fix to the double `update_report()` call assumes that `update_report()` will only be called if `pending.data` changes. That seems like a dangerous assumption.

thehunmonkgroup:
I'm going to upgrade my last comment, it's not a dangerous assumption, it's a wrong assumption ;)

Easy case in point: simply switching from one report to another does not update `pending.data`.

So your current solution to that problem will not work. Passing `pipe=True` as you are in the initial call to `update_report` is really a hack as well, and one I'm not will to have sprinkled throughout the code, as it would be highly likely to introduce further bugs because it's a non-obvious thing to consider.

Given the performance implications of calling `update_report()` twice for every write to `pending.data`, this PR will be on hold until that issue is addressed.

jspricke:
> I'm actually not crazy about the way the `update_report()` signature is altered. Isn't there a way we can do this and preserve report as the first arg? My preference would be to not change that function at all, but if that's not possible, then at least let's preserve existing argument order.

Fixed in the new version.

> I'm going to upgrade my last comment, it's not a dangerous assumption, it's a wrong assumption ;)
> 
> Easy case in point: simply switching from one report to another does not update `pending.data`.
> 
> So your current solution to that problem will not work. Passing `pipe=True` as you are in the initial call to `update_report` is really a hack as well, and one I'm not will to have sprinkled throughout the code, as it would be highly likely to introduce further bugs because it's a non-obvious thing to consider.
> 
> Given the performance implications of calling `update_report()` twice for every write to `pending.data`, this PR will be on hold until that issue is addressed.

Fixed in the new version.

jspricke:
> And in `on_modified` you're just writing garbage to the pipe, because it's not needed?

yes, that's the interface urwid provides.

thehunmonkgroup:
First, I want to say that I appreciate all the work you've put into this PR, including the more recent responses to my questions, and the code adjustments based on the concerns.

However, the last round of changes has me feeling pretty confident that we're headed in the wrong direction with this feature. While it cleaned up the calls to `update_report()`, it's also clearly exposed all the places where it's necessary to insert code to account for the double update issue, and the smell of incorrect architecture is now pretty strong ;)

The only reasonable thought I've had is to create some kind of simple socket listener, that other programs could write to in order to control VIT. While that doesn't really solve the double update problem if you're *just* watching for a change to `pending.data`, it *does* provide the opportunity for whatever outside process makes an outside update to send a signal to VIT. This seems like the correct location of responsibility: the outside process knows when it makes a change, and it signals VIT.

I've given little thought to the architecture of said socket interface, or the security implications, and there would need to be thought put into both if we went that route.

thehunmonkgroup:
With a little more thought, signals might be the way to go.

It seems completely appropriate to hook `SIGHUP` or `SIGUSR1` to the `refresh()` method, based on other uses I've seen of signals.

The signal interface is super simple, and I see no security concerns.

Check out the below patch which gets the `SIGUSR1` properly caught in the application code. However, there's an unfortunate issue at this point: calling `self.refresh()` there is *not* refreshing the interface. That makes zero sense to me, as the debug code will clearly illustrate, I've got access to a valid `Application` object. I think if we can solve that mystery, this would be the way to go.

```diff
diff --git a/vit/application.py b/vit/application.py
index 898eb86..dd5eb05 100644
--- a/vit/application.py
+++ b/vit/application.py
@@ -8,6 +8,7 @@ import subprocess
 import re
 import time
 import copy
+import signal
 from inspect import isfunction
 from functools import reduce
 
@@ -69,6 +70,7 @@ class MainFrame(urwid.Frame):
 
 class Application():
     def __init__(self, option, filters):
+        self.setup_signal_listeners()
         self.extra_filters = filters
         self.loader = Loader()
         self.load_early_config()
@@ -77,6 +79,15 @@ class Application():
         self.refresh(False)
         self.loop.run()
 
+    def setup_signal_listeners(self):
+        def sigusr1_handler(signum, frame):
+            import debug
+            debug.file("BEFORE UPDATE")
+            debug.file(self)
+            self.refresh()
+            debug.file("AFTER UPDATE")
+        signal.signal(signal.SIGUSR1, sigusr1_handler)
+
     def load_early_config(self):
         self.config = ConfigParser(self.loader)
         self.task_config = TaskParser(self.config)
@@ -917,6 +928,8 @@ class Application():
             raise RuntimeError("Error retrieving completed tasks: %s" % stderr)
 
     def refresh(self, load_early_config=True):
+        import debug
+        debug.file("BEGIN REFRESH")
         self.bootstrap(load_early_config)
         self.build_main_widget()
         # NOTE: Don't see any other way to clear the old palette.
@@ -924,6 +937,7 @@ class Application():
         self.loop.screen.register_palette(self.theme)
         self.loop.screen.clear()
         self.loop.widget = self.widget
+        debug.file("END REFRESH")
 
     def update_report(self, report=None):
         start = time.time()
```

jspricke:
>First, I want to say that I appreciate all the work you've put into this PR, including the more recent responses to my questions, and the code adjustments based on the concerns.

Thanks as well for being responsive and giving constructive feedback.
Given how long this PR takes, I would be open to have a video call to 
discuss details.

>It seems completely appropriate to hook `SIGHUP` or `SIGUSR1` to the `refresh()` method, based on other uses I've seen of signals.

I agree that signals and sockets are a way to trigger a refresh from 
the outside but it will not get you the features discussed here and, in 
addition, be a much more brittle interface with more assumptions and 
code. I will discuss this with signals but sockets basically have the 
same problems.

Signals are a way to communicate between processes. For the vit reload 
to work this would mean that every process modifying the taskwarrior DB 
would need to send a signal to all vit processes¹. Let's assume that only 
the task command modifies the DB², one way to send the signal would be 
via taskwarrior hooks.

The first problem would be to identify all relevant vit instances. We 
could get a process list from the kernel and trigger all those with vit 
in the command line³. But that would probably trigger to much and 
probably even unrelated processes, like an editor with the/a vit file 
opened. Given that SIGUSR1 has no clear semantics and can trigger all 
sorts of actions in a process, I'm not sure we want that and it would be 
quiet resource heavy.

The second problem would be the double update, as you identified 
already. Again with taskwarrior somehow triggering the signal, vit would 
need some way to inform taskwarrior that it should not trigger the 
signal in that case. This is basically what skip_watchdog in the current 
implementation does.

>Check out the below patch which gets the `SIGUSR1` properly caught in the application code. However, there's an unfortunate issue at this point: calling `self.refresh()` there is *not* refreshing the interface. That makes zero sense to me, as the debug code will clearly illustrate, I've got access to a valid `Application` object. I think if we can solve that mystery, this would be the way to go.

You want the signal to get delivered to the urwid loop and that would 
probably be through the same pipe interface as in this PR.

So overall we would get pretty much the same code as in this PR plus 
additional code elsewhere.

Btw. this is actually not surprising as the notify interface of the 
kernel is pretty much a signal tailored exactly to our use case and 
the Linux kernel uses a file descriptor for it. Also it gives you a much 
more fine grained control over the signal with clear semantics.

¹: My use case is that vit should always show me the state of the 
taskwarrior DB, i.e. what I get by calling task on the command line.

²: Given that the DB is text, there are probably other tools directly 
modifying it so all of them would need to send a signal to all vit 
instances.

³: Not that those with vit in the name would not be enough as on my 
computer they identify as "/usr/bin/python3 /usr/bin/vit" and the path 
could be arbitrary.

thehunmonkgroup:
I've added basic signal handling to VIT, see https://github.com/vit-project/vit/commit/10c67be4be35d9251528ce944c5120403adba2b0 for details.

I believe this provides the basic mechanism to allow outside processes to update VIT's interface.

I'll address some of your other concerns in another post.

thehunmonkgroup:
VIT's architecture, at least in terms of the interface, is request/response, similar to a non-AJAX/websockets load of a web page. I think this is a sensible architecture for VIT's general use case.

This PR in its current form is shoehorning in a feature that does not play well with that architecture, as is evidenced by all the code added to address the double update issue. What initially looked like a small, contained feature, is now another layer of consideration in maintaining the codebase, and a non-intuitive one at that -- I'm simply not willing to go in that direction.

As mentioned above, signals are a reasonable way for an outside process to communicate with VIT, and an easy, clean implementation.

> The first problem would be to identify all relevant vit instances. We
> could get a process list from the kernel and trigger all those with vit
> in the command line³. But that would probably trigger to much and
> probably even unrelated processes, like an editor with the/a vit file
> opened.

IMO this is overstated, and I believe there are simple ways to largely work around this issue. As an easy example, this would perfectly target all pids of all running VIT processes for my user:

```
[:~] $ ps aux | grep vit | grep -v grep
hunmonk  1652650  0.0  0.0  38840 21000 pts/0    S<+  11:09   0:00 /home/hunmonk/.pyenv/versions/3.9.11/bin/python vit/command_line.py
hunmonk  1653760  0.0  0.0  38848 21004 pts/5    S<+  11:11   0:00 /home/hunmonk/.pyenv/versions/3.9.11/bin/python vit/command_line.py
[:~] $ pgrep -U $UID -f "/home/hunmonk/.pyenv/versions/.*/bin/python vit/command_line.py"
1652650
1653760
```

Sure that's specific to my install, but I feel pretty confident that any user could write a small wrapper script to accurately pull their relevant PIDs.

I'm pretty sure it would also be trivial to add a `pid_dir` config directive, that, if enabled, would write the PID of a running VIT process to that directory in a [process id].pid file, and clean it up on exit. Most installations should support `/var/run/user/$UID`, or it can be written to `/tmp` if not, so that any stray PID files from a crash are cleaned up on next boot.

> The second problem would be the double update, as you identified
> already. Again with taskwarrior somehow triggering the signal, vit would
> need some way to inform taskwarrior that it should not trigger the
> signal in that case.

The double update issue is particular to the how the problem is being solved, and I believe it's a reasonable consideration that if an outside process is going to proactively trigger an update of VIT's interface, then it should know when it's the one doing the updating. For any automated scripts this should be fine, you just add a call to the wrapper script at the end that sends the SIGUSR1 to all running VIT processes. If you want to use a TaskWarrior hook, then in the simplest case you'd have to deal with the double update issue, but I'm sure there are ways to address that in the hook script. I haven't tested this, but I'd guess aliasing `task` to something like `IS_TASK_CLI=1 task`, and then looking for that environment variable in the hook would work.

I suppose the main counterargument to what I've said above it something like "but this is a lot of overhead for a user, when it could all be baked into VIT and just work". I would basically agree with that viewpoint, and add that IMO, it's completely appropriate to ask the user to do some work if they're trying to do something non-standard ;)

If we end up adding the `pid_dir` feature, you should have all the tools you need to wire up what you want.

Finally, I do have some curiosity as to how taskwarrior-tui handles this, but I do not have the time to investigate.

thehunmonkgroup:
https://github.com/vit-project/vit/commit/ecfa35c3778a5c61e46ce37f1db055eb72790fd0 adds a `pid_dir` option to VIT, which activates basic pid file management for VIT. It also includes a `--list-pids` command line argument, which returns all the pids in current pid files in `pid_dir`.

https://github.com/vit-project/vit/blob/2.x/scripts/vit-external-refresh.sh is an example script that illustrates the new feature in action, leveraging the new signal handling to auto refresh all running VIT instances.

I really don't believe anything else is needed here. Yes, if you want to use TaskWarrior hooks to trigger refresh you'll have to figure out a solution for the double update problem, or just deal with it, but as I mentioned before I believe this is a problem for the external program to solve.

thehunmonkgroup:
VIT now injects `IS_VIT_INSTANCE=1` into the environment during startup, and this is available to all hooks invoked by `task` via VIT.

Check out https://github.com/vit-project/vit/blob/2.x/scripts/hooks/on-exit-refresh-vit.sh for a TaskWarrior hook that intelligently refreshes all VIT interfaces, *except* when the hook is invoked during a `task` execution via VIT.

thehunmonkgroup:
I thought of another clear case where watching the `pending.data` file and refreshing would work poorly: imagine a script that modifies 100's of tasks automatically over the course of 5 minutes or so -- that would be maddening to try and work with VIT during that script.

But with the script in charge of when it refreshes VIT, it can do it at the end, once.

At this point I'm fully convinced that I've added all the necessary elements for a user to easily string together an auto refreshing VIT interface if they so desire, without adding any non-intuitive layers to the code, and it's flexible enough to handle all the use cases I've been exposed to.

Therefore, I'm closing this PR.

It would be great if somebody could open another PR to provide some user doc for the tools and features I've written in the last few days, and maybe provide a basic example setup so any user can configure their VIT install to have auto refreshing capability.

jspricke:
Thanks for working on this! As commented before your proposal does not work for me.
I will continue to maintain my patch, if someone is interested.

thehunmonkgroup:
Documentation for the auto refresh features has been added here: https://github.com/vit-project/vit/blob/2.x/CUSTOMIZE.md#auto-refreshing-vits-interface

PR #192, opened 2019-08-03
handle reports w/o columns info. "timesheets" for me doesn't.

dtzWill:
Might only be needed with latest taskwarrior 2.5.2, not sure, and wasn't sure if this was best way to handle it.

If nothing else, heads-up in case you encounter this down the road! :)

Thanks for all your work, BTW! I think the revived vit might be a key player in my return to taskwarrior goodness!

### DIFF
--- a/vit/config_parser.py
+++ b/vit/config_parser.py
@@ -343,6 +343,8 @@ class TaskParser(object):
         return primary_sort[0] == 'project' and primary_sort[1] == 'ascending'
 
     def get_column_index(self, report_name, column):
+        if not 'columns' in self.reports[report_name]:
+            return None
         return self.reports[report_name]['columns'].index(column) if column in self.reports[report_name]['columns'] else None
 
     def get_column_label(self, report_name, column):


thehunmonkgroup:
I'm not quite clear how a report can't have any columns, what's the name of the report you're trying to view? Is this a custom report?

thehunmonkgroup:
OK, I got 2.5.2 installed to test this myself. As reported, the ```timesheet``` report has no columns:

```sh
task _show | grep report.timesheet
report.timesheet.filter=(+PENDING and start.after:now-4wks) or (+COMPLETED and end.after:now-4wks)
```
However, I'm unclear if this is a bug in the development version of TaskWarrior, or some new kind of approach to a report that allows zero columns. I can't really fix this issue until I confirm with the TaskWarrior maintainer what's going on there.

dtzWill:
Thank you for investigating! Sorry for not responding sooner, apparently missed your message? :(

I'm also unsure how to best handle this, but FWIW it does appear to be limited to the `timesheet` "report".  Even so, waiting until there's a clearer picture of things sounds like the way to go.  In the meantime perhaps the PR will help lost taskwarrior users ("taskwarriors"? :grin:) encountering this also.  Thanks again.

thehunmonkgroup:
After talking with the TaskWarrior maintainers, I decided to take a different approach to fix this issue.

See the commit message in https://github.com/scottkosty/vit/commit/bf4cb069228e12b26e436b120fb9938a2b6a1531 for a full explanation.

dtzWill:
Perfect, thank you!

shaicoleman:
Can a new version be published to fix this?

thehunmonkgroup:
Doesn't warrant a point release, as the fix addresses features in an unreleased version of TaskWarrior.

You're welcome to run a development version of VIT should you wish to have access to the fix.

YAMLcase:
taskwarrior 2.5.2 has now been released and the bug persists

thehunmonkgroup:
@yamlCase the bug was fixed in https://github.com/vit-project/vit/commit/bf4cb069228e12b26e436b120fb9938a2b6a1531

VIT will probably get a point release in the near future, until then you can run the development version, which is compatible with TaskWarrior 2.5.2: https://github.com/vit-project/vit/blob/2.x/DEVELOPMENT.md

YAMLcase:
Thank you.  I can confirm the 2.X branch has the fix and runs fine under python 3.8 on Ubuntu 20.08.

PR #194, opened 2019-08-04
[2.x] Correct import

sanjayankur31:
Caught this when I ran `vit --list-actions` here on Fedora. I think these are the only instances, but worth a check in an installed environment (outside the source tree).


thehunmonkgroup:
Nice catch, thanks!

I did a quick review and those seem to be the only places the change was needed.

PR #198, opened 2019-08-11
Add support for XDG Base Directory

jspricke:


thehunmonkgroup:
I'm inclined to not include this PR in the core code, for the following reasons:

1. I have no familiarity with XDG
1. It's been difficult to find evidence that this is in fact a widely used standard -- as far as I can tell it's a FreeDesktop spec that may or may not be used in different distros
1. It adds another dependency, and I'm trying to avoid adding unnecessary dependencies
1. The architecture already includes a standard method for customizing the config directory

This would probably be better suited as a plugin if/when the plugin architecture gets added.

jspricke:
* thehunmonkgroup <notifications@github.com> [2019-08-17 13:47]:
>1. It's been difficult to find evidence that this is in fact a widely used standard -- as far as I can tell it's a FreeDesktop spec that may or may not be used in different distros

There is a lot of software supporting it:

https://wiki.archlinux.org/index.php/XDG_Base_Directory#Support

(And it's not about distros)

>1. It adds another dependency, and I'm trying to avoid adding unnecessary dependencies

What about just moving `~/.vit` to `~/.config/vit` without the 
dependency? This could be done easily before the first release (but 
gets harder later).

thehunmonkgroup:
From what I'm seeing in the 'Partial' support section, vit can leverage XDG config already with the following:

```sh
 export VIT_DIR=${XDG_CONFIG_HOME}/vit
```

I'm not seeing why anything else is needed?

jspricke:
>I'm not seeing why anything else is needed?

The intention of the XDG specification is to unify the location of 
configuration files in `$HOME` in one common name space. Using 
`$VIT_DIR` surely works but requires manual work and is not obvious to 
the user.

Debian recommends following the spec:

https://wiki.debian.org/UpstreamGuide#User_home_directories

And Gnome has some more arguments:

https://wiki.gnome.org/Initiatives/GnomeGoals/XDGConfigFolders

thehunmonkgroup:
I've discussed this PR briefly with a TaskWarrior maintainer, just to get some perspective, and I'm still not convinced this belongs in vit core.

The suggestion I posted above is enough to give somebody XDG compat if they want it, and https://github.com/scottkosty/vit/blob/2.x/CUSTOMIZE.md clearly states how to customize the vit config directory.

I'll leave this open for now, and if/when the plugin architecture gets added to vit, we can revisit making this work as a plugin.

justinmayer:
I must concur with @jspricke that XDG config directory support would be a valuable addition to Vit. It is indeed a widely-used standard. My beloved [Fish shell](http://fishshell.com) is but one notable example.

While I think the benefits have already been made clear, the inevitable consequence of everyone dumping configuration into $HOME is an extremely cluttered and unkempt home directory.

To add to previous comments on this topic...

> It adds another dependency, and I'm trying to avoid adding unnecessary dependencies.

I think the tendency to avoid dependencies is understandable but usually misguided. In the vast majority of cases, adding functionality via third-party packages confers more benefits than costs.

> vit can leverage XDG config already with the following:
>
> `export VIT_DIR=${XDG_CONFIG_HOME}/vit`
>
> I'm not seeing why anything else is needed?

1. As @jspricke already noted, this requires a cumbersome extra step.

2. It exchanges one source of clutter (configuration files/directories in $HOME) for two: shell configuration and environment variables. The last thing we need are dozens of lines in `.bashrc` solely dedicated to specifying configuration directory paths for various tools, accomplishing little except to pollute environment variable namespace.

> I've discussed this PR briefly with a TaskWarrior maintainer, just to get some perspective

I've seen their [gruff stance](https://github.com/GothenburgBitFactory/taskwarrior/issues/2203) on this matter, which I think is unfortunate and misguided. Thankfully, those who work on sister project TimeWarrior, under the same organization, do not share this view. [They see the value in supporting XDG home configuration.](https://github.com/GothenburgBitFactory/timewarrior/issues/207)

In summary, I think this pull request should be merged in its current form. But regardless of the specific implementation, it seems to me a fairly straightforward way to handle this in Vit core is to use `~/.vit` if it exists, and if not, use `${XDG_CONFIG_HOME}/vit` if it exists. What would be the harm in that?

thehunmonkgroup:
@jspricke @justinmayer I thank you for your efforts and clarifications, and appreciate the validity of some of your arguments in favor of XDG support. I'm pretty on the fence with this, but still coming down on the side of not merging.

As far as I can tell, this spec has been around for at least 10 years, and while quite a few apps support it, there are also still plenty that don't -- which doesn't exactly make the argument that it's a slam dunk ;)

Also, I've never personally had much of a problem with the organization of my ```$HOME``` directory, so even the valid arguments don't seem all that weighty to me. I can appreciate that others may have a different perspective.

Let's leave the PR open though, I do think we may revisit it someday.

00dani:
I'd like to bring this PR up again, because it turns out the upcoming 2.6.0 release of Taskwarrior [will be providing XDG Base Directory support](https://github.com/GothenburgBitFactory/taskwarrior/pull/2316) - it's scheduled to be released at the end of this month. So it seems all the more appropriate that a Taskwarrior wrapper like VIT provide automatic XDG support, for parity with Taskwarrior itself.

thehunmonkgroup:
This is on my list to look at, but it will be a few months, as I'm quite busy.

thehunmonkgroup:
OK, I've had a chance to review https://github.com/GothenburgBitFactory/taskwarrior/pull/2316 and the approach there seems sensible, and one I'd like to follow.

So for this PR to get in:

1. Ditch the dependency on `pyxdg`
2. Change the priority of the config file location to be:
    1. `${VIT_DIR}`
    2. `DEFAULT_VIT_DIR` (`${HOME}/.vit` as defined)
    3. `${XDG_CONFIG_HOME}/vit`

I agree w/ the Taskwarrior maintainer that we do not want to change the default location of the configuration, but we do want to provide a painless experience for those users wishing to leverage the XDG spec.

jspricke:
@thehunmonkgroup thanks for your review. I've updated this according to your comments.

thehunmonkgroup:
We're close. I understand the logic of the XDG search code, but it feels a bit heavy to jam that into the `__init__` method. Can we at least break that out into another method that `__init__` calls?

Also, there's a logic bug related to `VIT_DIR` in that `__init__` method, so breaking out the XDG code into a separate method will make that a bit easier to fix.

jspricke:
>Can we at least break that out into another method that `__init__` calls?

done.

thehunmonkgroup:
Thanks for everyone's work and persistence on this issue.

Issue #190, opened 2019-07-27
MAX_COLUMN_WIDTH = 60 is to small


jspricke:
For me this results in line breaks in the descriptions, and lot's of space to the right of it. So I patched it to 600. I see two options, either make it a configuration option, or compute use shutil.get_terminal_size and adopt the line length to the terminal. What would you prefer (I can probably come up with code for both of it)?

Thanks for porting vit to Python!

thehunmonkgroup:
Yeah, the column width calculation code needs more work. I'm already using the screen width and dynamically reducing wider columns, see resize_columns() in task_list.py

I'd be open to more optimizations to that code, although I recall it being fairly tricky to find an algorithm that worked well across a wide range of possible report/column scenarios -- you can't simply steal width from every column to make everything fit.

I think ideally there would be no need for MAX_COLUMN_WIDTH.

sanjayankur31:
Could there be a configuration option that allows setting percentages for the column widths maybe? Let there be "sane defaults", but it'll also let users tweak things easily?

thehunmonkgroup:
@sanjayankur31 pretty sure that would quickly become too complicated, as there are many reports, each with their own column configurations.

sanjayankur31:
Hrm, I wonder how `task` does it. :thinking:

thehunmonkgroup:
```task``` has the advantage of knowing that it will only have to display N rows of a report. We don't have that luxury, because the user can scroll down. I briefly reviewed ```task```'s strategy, and it seemed too processor heavy for VIT's use case.

sanjayankur31:
Ah, right---it's quite a complex problem then. Maybe just let `MAX_COLUMN_WIDTH` be configurable in the meantime and let users set it?

thehunmonkgroup:
I'm a bit resistant to add a config option for max column width, I'd rather see the algorithm improved and eliminate that idea entirely.

I'm hoping @jspricke will take a crack at that.

thehunmonkgroup:
Resolved by #199

PR #199, opened 2019-08-15
Remove MAX_COLUMN_WIDTH (Closes: #190)

jspricke:
The resize_columns() handles small terminals just fine.

### DIFF
--- a/vit/task_list.py
+++ b/vit/task_list.py
@@ -14,7 +14,6 @@ from vit.base_list_box import BaseListBox
 from vit.list_batcher import ListBatcher
 from vit.formatter.project import Project as ProjectFormatter
 
-MAX_COLUMN_WIDTH = 60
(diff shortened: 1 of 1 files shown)


jspricke:
For me this works, feel free to send problems (screen shots..) if not and I will have an other look.

thehunmonkgroup:
Yes, this makes sense, as I added the column reduction code much after the initial MAX_COLUMN_WIDTH hack.

The code works fine, but I'd prefer to keep the ```update_column_width()``` function -- it's cleaner, more descriptive, and gives a place to put other adjustments in the future. Could you please refactor your patch and push a rebased version?

jspricke:
>The code works fine, but I'd prefer to keep the ```update_column_width()``` function -- it's cleaner, more descriptive, and gives a place to put other adjustments in the future. Could you please refactor your patch and push a rebased version?

Sure, done.

PR #212, opened 2019-11-03
fix #176: Readline edit shortcuts for command bar

### LINKED ISSUE #176: Readline edit shortcuts for command bar
VIT 2.x does not yet have a full compliment of [Readline](https://tiswww.case.edu/php/chet/readline/rltop.html) shortcuts. Currently only 'go to line beginning' and 'go to line end' are implemented.

nfraprado:
New available readline commands are: Ctrl+b, Ctrl+f, Ctrl+h, Ctrl+d,
Ctrl+t, Ctrl+u, Ctrl+k, Meta+b, Meta+f, Ctrl+w, Meta+d.

These changes were based on [urwid readline](https://github.com/rr-/urwid_readline) where possible.
The code was adapted in places where urwid readline acted differently than expected from readline, e.g. in urwid, Meta+f jumped to the start of the next word, instead of the end of the current word.

Also, explicit reference to internal variables as used in urwid readline were exchanged by setter functions and normal variables (e.g. instead of using `self._edit_pos`, use `self.edit_pos` to get and `self.set_edit_pos()` to set). I'd really like feedback on this as I'm not sure what is the right thing to do here.

### DIFF
--- a/vit/readline.py
+++ b/vit/readline.py
@@ -1,0 +1,133 @@
+import string
+import re
+
+class Readline(object):
+    def __init__(self, edit_obj):
+        self.edit_obj = edit_obj
+        word_chars = string.ascii_letters + string.digits + "_"
+        self._word_regex1 = re.compile(
+            "([%s]+)" % "|".join(re.escape(ch) for ch in word_chars)
+        )
+        self._word_regex2 = re.compile(
+            "([^%s]+)" % "|".join(re.escape(ch) for ch in word_chars)
+        )
+
+    def keys(self):
+        return ('ctrl p', 'ctrl n', 'ctrl a', 'ctrl e', 'ctrl b', 'ctrl f',
+                'ctrl h', 'ctrl d', 'ctrl t', 'ctrl u', 'ctrl k', 'meta b',
+                'meta f', 'ctrl w', 'meta d')
+
+    def keypress(self, key):
+        # Move to the previous line.
+        if key in ('ctrl p',):
+            text = self.edit_obj.history.previous(self.edit_obj.metadata['history'])
+            if text != False:
+                self.edit_obj.set_edit_text(text)
+            return None
+        # Move to the next line.
+        elif key in ('ctrl n',):
+            text = self.edit_obj.history.next(self.edit_obj.metadata['history'])
+            if text != False:
+                self.edit_obj.set_edit_text(text)
+            return None
+        # Jump to the beginning of the line.
+        elif key in ('ctrl a',):
+            self.edit_obj.set_edit_pos(0)
+            return None
+        # Jump to the end of the line.
+        elif key in ('ctrl e',):
+            self.edit_obj.set_edit_pos(len(self.edit_obj.get_edit_text()))
+            return None
+        # Jump backward one character.
+        elif key in ('ctrl b',):
+            self.edit_obj.set_edit_pos(self.edit_obj.edit_pos - 1)
+            return None
+        # Jump forward one character.
+        elif key in ('ctrl f',):
+            self.edit_obj.set_edit_pos(self.edit_obj.edit_pos + 1)
+            return None
+        # Delete previous character.
+        elif key in ('ctrl h',):
+            if self.edit_obj.edit_pos > 0:
+                self.edit_obj.set_edit_pos(self.edit_obj.edit_pos - 1)
+                self.edit_obj.set_edit_text(
+                    self.edit_obj.get_edit_text()[0 : self.edit_obj.edit_pos]
+                    + self.edit_obj.get_edit_text()[self.edit_obj.edit_pos + 1 :])
+            return None
+        # Delete next character.
+        elif key in ('ctrl d',):
+            if self.edit_obj.edit_pos < len(self.edit_obj.get_edit_text()):
+                edit_pos = self.edit_obj.edit_pos
+                self.edit_obj.set_edit_text(
+                    self.edit_obj.get_edit_text()[0 : self.edit_obj.edit_pos] +
+                    self.edit_obj.get_edit_text()[self.edit_obj.edit_pos + 1 :])
+                self.edit_obj.set_edit_pos(edit_pos)
+            return None
+        # Transpose characters.
+        elif key in ('ctrl t',):
+            # Can't transpose if there are less than 2 chars
+            if len(self.edit_obj.get_edit_text()) < 2:
+                return None
+            self.edit_obj.set_edit_pos(max(2, self.edit_obj.edit_pos + 1))
+            new_edit_pos = self.edit_obj.edit_pos
+
+            edit_text = self.edit_obj.get_edit_text()
+            self.edit_obj.set_edit_text(
+                        edit_text[: new_edit_pos - 2] +
+                        edit_text[new_edit_pos - 1] +
+                        edit_text[new_edit_pos - 2] +
+                        edit_text[new_edit_pos :])
+            self.edit_obj.set_edit_pos(new_edit_pos)
+            return None
+        # Delete backwards to the beginning of the line.
+        elif key in ('ctrl u',):
+            self.edit_obj.set_edit_text(self.edit_obj.get_edit_text()[self.edit_obj.edit_pos :])
+            self.edit_obj.set_edit_pos(0)
+            return None
+        # Delete forwards to the end of the line.
+        elif key in ('ctrl k',):
+            self.edit_obj.set_edit_text(self.edit_obj.get_edit_text()[: self.edit_obj.edit_pos])
+            return None
+        # Jump backward one word.
+        elif key in ('meta b',):
+            self.jump_backward_word()
+            return None
+        # Jump forward one word.
+        elif key in ('meta f',):
+            self.jump_forward_word()
+            return None
+        # Delete backwards to the beginning of the current word.
+        elif key in ('ctrl w',):
+            old_edit_pos = self.edit_obj.edit_pos
+            self.jump_backward_word()
+            new_edit_pos = self.edit_obj.edit_pos
+            self.edit_obj.set_edit_text(self.edit_obj.edit_text[: new_edit_pos]
+                + self.edit_obj.edit_text[old_edit_pos :])
+            self.edit_obj.set_edit_pos(new_edit_pos)
+            return None
+        # Delete forwards to the end of the current word.
+        elif key in ('meta d',):
+            edit_pos = self.edit_obj.edit_pos
+            self.jump_forward_word()
+            self.edit_obj.set_edit_text(self.edit_obj.edit_text[: edit_pos]
+                + self.edit_obj.edit_text[self.edit_obj.edit_pos :])
+            self.edit_obj.set_edit_pos(edit_pos)
+            return None
+
+    def jump_backward_word(self):
+        for match in self._word_regex1.finditer(
+            self.edit_obj.edit_text[: self.edit_obj.edit_pos][::-1]
+        ):
+            self.edit_obj.set_edit_pos(self.edit_obj.edit_pos - match.end(1))
+            return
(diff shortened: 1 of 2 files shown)


thehunmonkgroup:
Haven't had time to test in depth yet, but here are my thoughts from first review:

Code generally looks good.

Yes, I think you've taken the right approach using the getter/setters instead of internal variables,

I'm a little uncomfortable how giant the conditional is getting in ```keypress()```. I think it would be better to make this a module, perhaps used something like this:

In ```command_bar.py```
```python
from vit.readline import Readline

# ... in __init__()
self.readline = Readline(self)

# ... in keypress()
elif key in Readline.keys():
  return Readline.keypress(key)
```

In ```vit/readline.py```
```python
class Readline(object):

  def __init__(self, edit_obj):
    self.edit_obj = edit_obj

  def keys(self):
    return ('ctrl a', 'ctrl b') # and all the other shortcuts

  def keypress(self, key):
    if key in ('ctrl a'):
      # etc
```

Advantages:

1. Clean separation of functionality
2. Possibly re-usable in other future places besides the command bar
3. Easier to unit test

Thoughts?

nfraprado:
Sounds good to me!
Should I already make these changes, test them and push it here or would you like to wait for others to give feedback?

thehunmonkgroup:
Yep, go ahead and make the changes -- on a project this small, two is a quorum ;)

nfraprado:
I'm in the process of making the change you proposed and I have question: would you consider 'ctrl p'/Up and 'ctrl n'/Down as part of readline or not? (Meaning, should I move them too?)

Also, I don't have much experience in contributing through GitHub, so I would like your opinion here. To push these changes, I was thinking about basically reverting my first commit (apart from removing that comment) and making the change you said in the same commit. Is that okay or is there a cleaner way to do it?

nfraprado:
Just to clarify regarding the commits: there would be 2 commits, the current one, plus another one making the changes you proposed. (As opposed to creating a new branch so that everything is done in a single commit). I ask this because this second commit would be removing a lot of the lines added in the first one.

thehunmonkgroup:
Looks like ```Ctrl+p``` and ```Ctrl+n``` are in the official readline, so how about we move those to the module, and leave ```up``` and ```down``` as custom key presses specific to the command bar?

Glancing at my example code again, I'm realizing that in order for future modules to leverage the abstraction, they'll have to conform to the way ```edit_obj``` is used in readline module. There might be a cleaner approach long term, but this is still a much better abstraction, so let's continue forward.

I'd just go ahead and add commits to the existing branch. When we're ready to merge, we can decide if we want to squash the commits in your PR into one commit or not: https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-request-merges#squash-and-merge-your-pull-request-commits

nfraprado:
In that case, you mean having duplicated code to treat going backwards and forwards in command history, one for `up`/`down` in `command_bar.py` and another one for `Ctrl+p`/`Ctrl+n` in `readline.py`? 
Or rather only move the implementation from `command_bar.py` to `readline.py` and make `up` in `command_bar.py` call `keypress('ctrl p')` in `readline.py` and `down` call `keypress('ctrl n')`?

thehunmonkgroup:
Well, the thing is I'm not sure if ```up``` and ```down``` are standard Readline key commands, are they?

If so, then it makes sense to put them in our Readline module. If not, then a cleaner separation would be to *not* put them in our Readline module, and instead have separate handling in ```command_bar.py``` -- which I guess could just call the ```Ctrl+p``` and ```Ctrl+n``` code in Readline to reduce code duplication.

nfraprado:
So, I checked the info page for readline and it seems that `up` and `down` indeed aren't part of readline, so I kept them in `command_bar.py` while moving `Ctrp+p` and `Ctrl+n` to `readline.py`. To reduce code duplication I made `up` and `down` only call `Ctrl+p` and `Ctrl+n`, respectively.

One issue I had is that the class `CommandBarHistory` inside `command_bar.py` seemed specific to the command bar, so I kept it in `command_bar.py` instead of moving it to `readline.py` even though it implements the history of commands used by readline. This means that `Ctrl+n` and `Ctrl+p` actually access the history object from `command_bar.py`, which means that in order to use the readline module, future components need not only to conform to the way the Edit object is used, but also need to contain a history object that conforms to the way it is used.

thehunmonkgroup:
Oh yeah, this is looking much better :) Nice work!

It was precisely the history stuff I was thinking of in my comment about there being a cleaner abstraction. I did review the ```CommandBarHistory``` class, and it's actually pretty generic -- ```add```, ```previous```, ```next```, ```cleanup```, across multiple commands. I think it would be pretty easy to abstract that, but I'll file another issue for it -- we'll commit this as the first step.

Code structure looks good on first quick read, I still need to do a deeper dive and run some tests.

vit/readline.py
```diff
@@ -0,0 +1,118 @@
...
+                edit_pos = self.edit_obj.edit_pos
+                self.edit_obj.set_edit_text(
+                    self.edit_obj.get_edit_text()[0 : self.edit_obj.edit_pos] +
+                    self.edit_obj.get_edit_text()[self.edit_obj.edit_pos + 1 :])
+                self.edit_obj.set_edit_pos(edit_pos)
+            return None
+        elif key in ('ctrl t',):
+            # Can't transpose if there are less than 2 chars
+            if len(self.edit_obj.get_edit_text()) < 2:
+                return
```
thehunmonkgroup: I think this should be ```return None```?

vit/readline.py
```diff
@@ -0,0 +1,118 @@
...
+            "([^%s]+)" % "|".join(re.escape(ch) for ch in word_chars)
+        )
+
+    def keys(self):
+        return ('ctrl p', 'ctrl n', 'ctrl a', 'ctrl e', 'ctrl b', 'ctrl f',
+                'ctrl h', 'delete', 'ctrl d', 'ctrl t', 'ctrl u', 'ctrl k',
+                'meta b', 'meta f', 'ctrl w', 'meta d')
+
+    def keypress(self, key):
+        if key in ('ctrl p',):
```
thehunmonkgroup, line 22: I think it would help code readability to add a comment above all these key conditionals, saying what Readline action it's for, e.g.:

```python
# Move to the previous line.
if key in ('ctrl p',):
   ...
```

thehunmonkgroup (changes requested):

vit/readline.py
```diff
@@ -0,0 +1,118 @@
...
+            self.edit_obj.set_edit_pos(self.edit_obj.edit_pos + 1)
+            return None
+        elif key in ('ctrl h',):
+            if self.edit_obj.edit_pos > 0:
+                self.edit_obj.set_edit_pos(self.edit_obj.edit_pos - 1)
+                self.edit_obj.set_edit_text(
+                    self.edit_obj.get_edit_text()[0 : self.edit_obj.edit_pos]
+                    + self.edit_obj.get_edit_text()[self.edit_obj.edit_pos + 1 :])
+            return None
+        elif key in ('delete', 'ctrl d'):
```
thehunmonkgroup: According to https://readline.kablamo.org/emacs.html ```delete``` and ```ctrl d``` have slightly different functionalities.

Also, I believe the ```delete``` key press is already handled just fine natively by Urwid, so I think we can simply remove that from both here and the ```keys()``` method.

thehunmonkgroup (changes requested):

vit/readline.py
nfraprado: Yeah, I don't think it makes a difference in the end, but I fixed the typo.

vit/readline.py
nfraprado: Of course. I just added the descriptions. I tried to be consistent in the wording, but since I'm not a native speaker, feel free to correct any grammar mistakes!

vit/readline.py
nfraprado: Right. Just removed `delete` from readline module.

vit/readline.py
```diff
@@ -55,7 +55,7 @@ def keypress(self, key):
                     + self.edit_obj.get_edit_text()[self.edit_obj.edit_pos + 1 :])
             return None
         # Delete next character.
-        elif key in ('delete', 'ctrl d'):
+        elif key in ('ctrl d'):
```
thehunmonkgroup: We need to have a comma after ```'ctrl d'```, otherwise I don't think Python will see that as a single item tuple. See the other conditionals checks in that section.

thehunmonkgroup:
This PR is *very* close, just need to:
1. Fix the missing comma I mentioned above
2. Merge 2.x and squash the commits in preparation for merging this branch back into 2.x

vit/readline.py
nfraprado: Yeah, sorry about that. Should have been testing these past changes.

Though now while testing I remembered why I added `delete` to readline like that. It seems that its implementation by urwid is kinda buggy, at least for me. Whenever I press `delete`, it deletes the next character but also positions the cursor at the end of the command line.

This also happens on the 2.x branch, so could you please also test it on your end?

thehunmonkgroup: I'm not seeing that behavior in my environment -- ```delete``` is functioning as expected, both deleting from the end of a line and from the middle of a line.

I would be pretty surprised if the Urwid library had such a bug -- properly deleting text in the middle of a string should be supported in the basic edit box functionality.

Can you try to reproduce the issue on your end? Maybe there's something specific you're doing that I didn't try.

nfraprado: I just tested a sample code using just urwid and it worked fine so it really isn't a bug in urwid.

Though I also tested using a clean vit from pip and a clean taskwarrior and vit config, and the issue does happen. So it doesn't have to do with the patch, nor with my config. But if you can't reproduce it then it should be something to do with my environment. Will have to test more.

The issue isn't anything specific. Everytime I press `delete` in any command bar inside vit, be it the prompt for typing a command or the `Add:` prompt, it deletes the next character but also positions the cursor at the end of the line. When I overrode the handling of `delete` with my readline module (as in this code snippet), the issue didn't happen.

thehunmonkgroup: How do you want to proceed?
1. Continue to investigate the bug in your local environment to see if you can isolate why it's happening, or...
2. Complete the PR without re-adding ```delete``` to the Readline shortcuts at this time

nfraprado: Okay, since this issue seems to be specific to my environment, it makes more sense to proceed this PR removing `delete` (option 2) and later I can investigate this problem further and open an issue for it.

nfraprado:
I'm not entirely familiar with the processes indicated in 2.

By merging 2.x you mean that I should merge the readline branch of my fork with the 2.x branch of my fork?

And regarding squashing the commits, I'm not really sure how I could do that on Github. I can't squash the commits using `git rebase -i` since it is already pushed to Github, right? Also, would you like it to be a single commit with all the changes?

nfraprado:
Just remembered you pointed to an article regarding squash and merge earlier in this thread. I gave it a read but from my understanding you're the one that would have to do the squashing since you have the write permissions on the repo.

thehunmonkgroup:
I was off a bit in my previous request. Here's basically how I think you'd do it:

```sh
git checkout 2.x
git pull https://github.com/scottkosty/vit.git
git checkout readline
git rebase -i 2.x
```
...then use the interactive rebase to mark everything ```fixup``` for everything but one commit. Optionally ```git commit --amend``` to touch up the commit message, then ```git push -f``` to force push your branch upstream.

Although you don't have write access to this repository, you have it to your PR branch, and therefore you can do a force push to have the rebased commits overwrite your older commit history.

Then I'll merge your adjusted branch.

Let me know if you need anything else.

nfraprado:
Okay, I did the squashing and the force-push. Let me know if I did it right, or what should be done.

thehunmonkgroup:
@nfraprado , excellent work on this, thanks again!

PR #214, opened 2019-11-15
Change background color of table stripes

thomasrebele:
I was overwhelmed when I tried the new version. The version 1.3 was quite restrained. However, the themes in 2.0 are just too much contrast. Can we reduce this to 10% black? I also added __pycache__ to .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .python-version
+__pycache__/
 
 # Setuptools build/distribution folder.
 /build/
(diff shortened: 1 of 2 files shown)


thehunmonkgroup:
I'm going to leave the current core themes as is -- however, you can certainly use a custom user theme to get the look you're wanting: https://github.com/scottkosty/vit/blob/2.x/CUSTOMIZE.md#themes

Would you mind submitting the .gitignore change in another PR? I'd like to keep the PRs to one per distinct change. Or if you'd rather not, I can handle adding it myself, just let me know.

PR #238, opened 2020-05-16
Fix crash when report has project column but no sort

### LINKED ISSUE #237: Vit crashes when report has project column but no sort
When a report is configured to have `project` as one of the columns but no `sort` is defined for the report, vit crashes with the following traceback:

```
  File "/usr/bin/vit", line 11, in <module>
    load_entry_point('vit', 'console_scripts', 'vit')()
  File "/home/nfraprado/code/contrib/vit/vit/command_line.py", line 5, in main
    Application(options, filters)
  File "/home/nfraprado/code/contrib/vit/vit/application.py", line 70, in __init__
    self.load_early_config()
  File "/home/nfraprado/code/contrib/vit/vit/application.py", line 79, in load_early_config
    self.reports = self.task_config.get_reports()
  File "/home/nfraprado/code/contrib/vit/vit/config_parser.py", line 342, in get_reports
    self.reports[report_name] = self.rectify_report(report_name, report)
  File "/home/nfraprado/code/contrib/vit/vit/config_parser.py", line 346, in rectify_report
    report['subproject_indentable'] = self.has_project_column(report_name) and self.has_primary_project_ascending_sort(report)
  File "/home/nfraprado/code/contrib/vit/vit/config_parser.py", line 353, in has_primary_project_ascending_sort
    primary_sort = report['sort'][0]
KeyError: 'sort'
```

To reproduce, simply generate a dummy install and add the following:
```
report.test.columns=id,project
report.test.labels=ID,Project
```
If a sort, even if empty, is added, the error is solved:
```
report.test.sort=
```

nfraprado:
Surround the `sort` key access from the report in a try block.

Fix #237 .

### DIFF
--- a/vit/config_parser.py
+++ b/vit/config_parser.py
@@ -350,7 +350,11 @@ class TaskParser(object):
         return self.get_column_index(report_name, 'project') is not None
 
     def has_primary_project_ascending_sort(self, report):
-        primary_sort = report['sort'][0]
+        try:
+            primary_sort = report['sort'][0]
+        except KeyError:
+            return False
+
         return primary_sort[0] == 'project' and primary_sort[1] == 'ascending'
 
     def get_column_index(self, report_name, column):


vit/config_parser.py
```diff
@@ -350,8 +350,12 @@ def has_project_column(self, report_name):
         return self.get_column_index(report_name, 'project') is not None
 
     def has_primary_project_ascending_sort(self, report):
-        primary_sort = report['sort'][0]
-        return primary_sort[0] == 'project' and primary_sort[1] == 'ascending'
+        try:
+            primary_sort = report['sort'][0]
+        except KeyError:
+            return False
+        else:
```
thehunmonkgroup: This ```else``` seems superfluous, given the ```return ``` above?

nfraprado: Yes, you're right. I'll remove it.

PR #243, opened 2020-06-17
fix typos

tbm:


thehunmonkgroup:
'denoting' -- good catch

It is forever my curse to spell the word 'occasionally' wrong -- two C's and one S? One C and two S's?

Thanks for the attention to detail.

scottkosty:
@thehunmonkgroup I don't think the "denoting" change was correct. I'm not familiar with GitHub review. Should I have done "Start review"? I think I messed it up by just commenting on a line of code.

thehunmonkgroup:
Well, I did some Google searching, and 'denoting' seems to be the associated verb. Are you seeing something else?

scottkosty:
We are talking about removing an annotation right? Taskwarrior [docs](https://taskwarrior.org/docs/commands/) list "denotate" as "Remove an annotation from a task".

thehunmonkgroup:
Have a look at this and see what you think: https://wikidiff.com/denotation/annotation

scottkosty:
I think I see your argument, which is that the verb "denote" shows up in the first definition on that link, but I don't think that any of the definitions on the link you gave refers to the concept of removing an annotation, or removing anything or undoing anything, which is what "denotate" attempts to refer to (see the commit where it was first introduced [here](https://github.com/GothenburgBitFactory/taskwarrior/commit/5e53226eb8ba0767a714efa5489a11525a05d2ba)). Let's take a look at the phrase in question for our discussion: "The method of denoting tasks has changed." After the change, it is now misleading. To "denote" is to indicate or to mark, so it seems the phrase is referring to how tasks are marked, instead of how the annotations of tasks are removed.

In any case, I'm fine with leaving "denote" in UPGRADE.md. I was really just curious about whether I did something wrong regarding the GitHub review since I don't have a lot of experience with reviewing pull requests on GitHub and using their review feature. I thought you did not see my comment so I wanted to learn how to do it correctly, but now I understand that you saw it and you just disagreed. No problem!

thehunmonkgroup:
'Denotate' is actually how Taskwarrior refers to the action in question, so we're kind of stuck with it.

Let's just do this: https://github.com/scottkosty/vit/commit/8d885c24dbbfef4d7585b6fcb68d38a7b4dda15a

scottkosty:
> Let's just do this: [8d885c2](https://github.com/scottkosty/vit/commit/8d885c24dbbfef4d7585b6fcb68d38a7b4dda15a)

That works, thanks.

PR #244, opened 2020-06-20
README.md: remove reference to 1.3

scottkosty:
@thehunmonkgroup I think we linked to 1.3 after we made 2.x the default branch, but I think the transition to 2.x is complete and smooth. Please merge if you agree.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,8 +2,6 @@
 
 Visual Interactive Taskwarrior full-screen terminal interface.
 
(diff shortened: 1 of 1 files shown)


thehunmonkgroup:
*Well...*

There are still plenty of distros out there that are installing 1.3.x packages (unfortunately). I think it's going to be that way for awhile.

My sense is that the home page should have some minimal reference so those folks can find the 1.3 branch if they really want to look at it?

scottkosty:
Ah good to know. I should have thought of distros but was only thinking about individual users. Good idea to wait.

PR #248, opened 2020-07-08
keybinding variable to retrieve any task attribute

thehunmonkgroup:
form: {TASK.[attribute]}, e.g. {TASK.project}

I'm not 100% sure I like the implementation, but it seems to work. Feedback appreciated.

### DIFF
--- a/CUSTOMIZE.md
+++ b/CUSTOMIZE.md
@@ -59,13 +59,51 @@ By default, VIT uses the ```keybinding/vi.ini``` configuration to provide Vi-sty
 
 To see the list of actions that can be mapped, execute ```vit --list-actions```.
 
-##### To override default keybindings:
+#### To override default keybindings:
 
 The ```[keybinding]``` section in ```config.ini``` overrides any core keybindings or keybindings that you place in the user ```keybinding``` directory. If you just want to make some small tweaks and/or add some macros, it's probably better to take this approach.
 
 The [config.sample.ini](https://github.com/scottkosty/vit/blob/2.x/vit/config/config.sample.ini) has many examples to illustrate how to customize keybindings for actions and add macros, check it out!
 
-##### To provide your own default keybindings:
+#### To provide your own custom variable replacements:
+
+VIT exposes a simple API to provide your own variable replacements in keybindings.
+
+Variables enclosed in curly brackets that VIT doesn't know about will be passed to your custom code,
+where you can match against the variable, and parse the string to extract metadata in the form of
+arguments that you pass to your custom replacement callback.
+
+To provide custom variable replacement:
+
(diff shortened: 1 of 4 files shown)


thehunmonkgroup:
OK, I'm *much* happier with this newer implementation, it's turned into a simple API that feeds a variable name to a match callback, which returns a list of arguments if the variable matches that particular replacement type. When the actual replacement happens, a  replacement callback is called and passed the list of arguments.

This allows complete flexibility to the replacement type to define how it's matched, and what arguments if any it wants to receive.

The reference implementation is ```{TASK_[attribute]}```, but others could easily be added. It would probably also be trivial to expose this to users so they can include their own custom variable replacement functionality.

thehunmonkgroup:
Turned out to be four lines of code to expose this API to users, and you can do some pretty fun stuff with it, check out CUSTOMIZE.md in the PR branch for an explanation.

As a example, I wrote up a quick custom variable replacement that displays local weather...

```python
# keybinding/keybinding.py
import re
import requests
from datetime import datetime, timezone

class Keybinding(object):

    def __init__(self):
        self.weather_replacement_match = re.compile('^WEATHER_(.+)$')
        # Get API key from:
        # https://rapidapi.com/community/api/open-weather-map?endpoint=53aa6041e4b00287471a2b62
        self.open_weather_api_key = 'you API key'

    def replacements(self):
        def _weather_match(variable):
            matches = re.match(self.weather_replacement_match, variable)
            if matches:
                city_state = matches.group(1)
                return [city_state]
        def _weather_replace(task, city_state):
            url = 'https://community-open-weather-map.p.rapidapi.com/weather'
            querystring = {
                'units': 'imperial',
                'q': city_state,
            }
            headers = {
                'x-rapidapi-host': 'community-open-weather-map.p.rapidapi.com',
                'x-rapidapi-key': self.open_weather_api_key,
            }
            response = requests.request('GET', url, headers=headers, params=querystring)
            json = response.json()

            desc = json['weather'][0]['description']
            temp = json['main']['temp']
            humidity = json['main']['humidity']
            sunrise = datetime.utcfromtimestamp(json['sys']['sunrise']).replace(tzinfo=timezone.utc).astimezone(tz=None).strftime('%I:%M%p')
            sunset = datetime.utcfromtimestamp(json['sys']['sunset']).replace(tzinfo=timezone.utc).astimezone(tz=None).strftime('%I:%M%p')

            return "Weather in %s: %s, temp %s, humidity %s, sunrise: %s, sunset: %s" % (city_state, desc, temp, humidity, sunrise, sunset)
        return [
            {
                'match_callback': _weather_match,
                'replacement_callback': _weather_replace,
            },
        ]

```

PR #263, opened 2020-08-05
catch re compile errors to prevent crashing

bookshelfdave:
vit crashes when you enter `/+foo`, as it's an invalid regex. I was originally trying to search for a tag when vit crashed. This PR catches the exception and shows a regex error.

I noticed a similar issue with `:help +foo`, hope you don't mind both changes in 1 PR.

Note: I don't see a test suite for the `vit.Application` class, and there's a lot of wiring needed to setup a test instance.

### DIFF
--- a/vit/help.py
+++ b/vit/help.py
@@ -104,17 +104,23 @@ class Help(object):
         ]
 
     def update(self, filter_args):
-        entries = self.filter_entries(filter_args)
-        self.listbox.reload_entries(entries)
+        try:
+            entries = self.filter_entries(filter_args)
+            self.listbox.reload_entries(entries)
+            self.error_widget.set_text('')
+        except ValueError as e:
+            self.error_widget.set_text(str(e))
+            self.listbox.reload_entries([])
         return self.widget
 
     def filter_entries(self, filter_args):
         if len(filter_args) > 0:
             try:
-                args_regex = re.compile('.*(%s).*' % '|'.join(filter_args))
+                rexp = '.*(%s).*' % '|'.join(filter_args)
+                args_regex = re.compile(rexp)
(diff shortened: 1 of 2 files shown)


vit/help.py
```diff
@@ -110,7 +110,10 @@ def update(self, filter_args):
 
     def filter_entries(self, filter_args):
         if len(filter_args) > 0:
-            args_regex = re.compile('.*(%s).*' % '|'.join(filter_args))
+            try:
+                args_regex = re.compile('.*(%s).*' % '|'.join(filter_args))
+            except:
```
thehunmonkgroup, line 122: Can we put a more specific exception in here? I prefer to catch errors more specifically so we get legitimate crashes on an unexpected error.

thehunmonkgroup (changes requested):

vit/help.py
bookshelfdave: yep, I left out an exception because I didn't see a widget to show an error message

thehunmonkgroup: Well there are two components here:
1. Displaying an error message to the user
2. Focusing the exception on the expected error

For number one, there is a facility in ```application.py``` to display error messages, as such:
```python
        except VitException as err:
            self.activate_message_bar(str(err), 'error')
            return
```

However, the application object that exposes that method is not available in the help. I'm not sure yet of a way to work that into the help without either a fairly large refactor, or an ugly hack, I'll have to think about it.

Number two we should be able to do better with though. What is the exact error type that is thrown when ```re.compile``` fails because of a bad regex? That's what we want for the exception line, something like:
```python
except TypeError:
```
...although I'm not sure what the right error type is in this case. The reasoning there is that we want VIT to crash for any error there except the expected one we're catching.

PR #267, opened 2020-08-22
Allow doing multiple task selection to modify them all at once

sefeng211:
Hi,

This patch adds two things

1) Add a new select command, which allows users to select tasks by pressing `v`. 
2) When using the modify command (`m`), when there are multiple tasks selected, it'll execute the command against all selected tasks. 

Let me know how you think about the patch and please feel free to reject it because I was basically just hacking around :)

### DIFF
--- a/vit/task_list.py
+++ b/vit/task_list.py
@@ -20,7 +20,7 @@ MARKER_COLUMN_NAME = 'markers'
 
 class TaskTable(object):
 
-    def __init__(self, config, task_config, formatter, screen, on_select=None, event=None, action_manager=None, request_reply=None, markers=None, draw_screen_callback=None):
+    def __init__(self, config, task_config, formatter, screen, on_select=None, event=None, action_manager=None, request_reply=None, markers=None, draw_screen_callback=None, selected_tasks=None):
         self.config = config
         self.task_config = task_config
         self.formatter = formatter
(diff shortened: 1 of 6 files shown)


vit/task_list.py
```diff
@@ -151,6 +154,12 @@ def update_focus_attr(self, attr, position=None):
             position = self.listbox.focus_position
         self.list_walker[position].row.set_attr_map({None: attr})
 
+    def update_selection(self):
+        selected_task_ids = set([task['id'] for task in self.selected_tasks])
+        for item in self.list_walker:
```
sefeng211, line 159: This function probably needs an optimization because it loops through all tasks, ideally we should only need to loop though the selected ones. 

I'd like to get some thoughts about this patch first before I dig into this :)

thehunmonkgroup:
Apologies for the long delay in my reply.

First, thanks for submitting the PR, it's a bold effort, and I appreciate somebody willing to do the work to make VIT better!

I have not fully vetted the patch yet, it touches a fair amount of code, and this change really needs to be executed cleanly or it could become a mess to maintain alongside other features.

Here are some initial observations after a brief look at the code, and playing with the functionality a bit:

1. The 'selected' row highlighting obscures the row highlighting for the focused task, which can easily get somebody lost when they are trying to scroll through a longer list of previously selected tasks. I'm not 100% confident that full row highlighting is the solution here. I have a suspicion that dynamically adding a special 'operations' column any time there are pending operations on a task might be both more flexible for the addition of future operations, and more visually integrated.
2. Doing a full ```self.update_task_table()``` for each selected item is a bit heavy, and would definitely operate poorly in situations where there was a big task list. It would be much more efficient to use an approach similar to how the focused task is handled. Although if we went with that 'operations' column idea I suggested in number one above, then doing things per row would be more complicated. Tricky.
3. We *definitely* want to handle modifying multiple tasks in one ```task``` operation, like so: ```task 2,3 modify +sometag```
4. The biggest weakness of the patch is that it only applies to the ```modify``` operation. I don't think this alone will be enough for me to accept the patch, as multi-select really should work across most other operations (project, tag, done, delete, etc).

sefeng211:
Thanks for the review!! They are helpful and indicated what needs to be done for this feature. 

Sounds like the first step should be `adding a special 'operations' column` so that this feature can rely on it. 

Closing this PR as I don't know if I'll make that happen.

PR #271, opened 2020-10-08
Add default keybindings for jumping to task number

nfraprado:
Since none of the numbers have default keybindings at the moment, and since apparently vit doesn't support prepending a command with a count like vim, assign default keybindings to each digit that opens an ex command with the digit.

This is just for convenience, so that the user that wants to jump to task 123 can just type `123`, instead of `:123`. Just one keystroke saved, but a free one.

This isn't really the `vim`-way to do things, this binding is actually the default in `mutt`, but since `vi.ini` seemed to be the only file for default keybindings, I added it here. We could add it to a different file if it is possible to source both at startup by default.

### DIFF
--- a/vit/keybinding/vi.ini
+++ b/vit/keybinding/vi.ini
@@ -48,6 +48,16 @@ M = {ACTION_LIST_SCREEN_MIDDLE}
 L = {ACTION_LIST_SCREEN_BOTTOM}
 C = {ACTION_LIST_FOCUS_VALIGN_CENTER}
 
+1 = :1
+2 = :2
+3 = :3
+4 = :4
+5 = :5
+6 = :6
+7 = :7
+8 = :8
+9 = :9
+
 [report]
 f = {ACTION_REPORT_FILTER}
 <Ctrl>l = {ACTION_REFRESH}


thehunmonkgroup:
I'm a little confused how this will work. Wouldn't those bindings only be valid for task IDs 1-9?

nfraprado:
No, sorry, I should have been more clear. There is no `<Enter>` at the end of the binding, so the user has to press Enter at the end, so it can be any number, not only single digits. The bindings aren't actually to jump to a task per se, but to enter the ex command with the digit pressed.

To jump to task 1: `1<Enter>`. To jump to task 10: `10<Enter>`. We are just eliminating the initial press of `:` that was needed before this commit.

thehunmonkgroup:
I understand the convenience of the feature, but I'm not sure this should be part of the default keybindings.

What do you think about adding this in the documentation as a way users can add this customization if they so desire?

nfraprado:
If you're sure this shouldn't be the default, we could add it in the documentation, but I still think this should be default.

I don't see any disadvantages. The keybindings are currently free, so they won't cause bad surprises for the users.

While taskwarrior operates by issuing commands, vit changes the paradigm to be focused on editing the currently selected task. Most commands work like this. So moving the cursor to the desired task before performing some action on it is central to vit's operation.
That's why I think moving to a task should as easy as possible. Not that pressing an additional `:` is that tiring, but given how often it is used, the fact that pressing a digit unambigously means the user wants to go to a task, and the fact that those keys are just free, I don't see a reason not to make this the default keybinding.

Unless you're thinking about adding support for prepending counts to commands like vim does, but I think visual selection is much more natural for vit for multiple editions at once.

thehunmonkgroup:
They are free, but there are not a lot of single key presses left free, and I'm hesitant to add any more single key presses without careful consideration.

My preference would be, as an initial step, to add it to the doc, and let it be that way for awhile -- let's see if any other great ideas come up for those keys.

scottkosty:
I haven't looked at the details, but I can also say I'm hesitant to support changing default keybindings without strong support. If we want to change them later, users understandably get frustrated (even if we tell them how to configure their rc file to get the old behavior back).

nfraprado:
I opened #272 adding these keybindings to the documentation instead. Feel free to close this PR.

PR #272, opened 2020-10-14
Add digit-jumping keybindings suggestion to CUSTOMIZE doc

nfraprado:
As discussed in #271, keybindings mapping digits to ex commands can be convenient to the users, but shouldn't be the default, and should rather be suggested through the documentation.

This adds a "Keybinding suggestions" section to the CUSTOMIZE documentation with a subsection suggesting and explaining these keybindings.

### DIFF
--- a/CUSTOMIZE.md
+++ b/CUSTOMIZE.md
@@ -110,3 +110,27 @@ class Keybinding(object):
 1. Create a ```keybinding``` directory in the user directory
 2. Copy over one of the core keybindings, and customize to your liking.
 3. Set the ```default_keybindings``` setting in ```config.ini``` to the name of the keybinding file you created, without the ```.ini``` extension. For example, if you created ```keybinding/strange.ini```, you would set ```default_keybindings = strange``` in ```config.ini```
+
+#### Keybinding suggestions
+
+##### Jumping with digits
+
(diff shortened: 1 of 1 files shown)


CUSTOMIZE.md
```diff
@@ -110,3 +110,27 @@ class Keybinding(object):
 1. Create a ```keybinding``` directory in the user directory
 2. Copy over one of the core keybindings, and customize to your liking.
 3. Set the ```default_keybindings``` setting in ```config.ini``` to the name of the keybinding file you created, without the ```.ini``` extension. For example, if you created ```keybinding/strange.ini```, you would set ```default_keybindings = strange``` in ```config.ini```
+
+#### Keybinding suggestions
+
+##### Jumping with digits
+
+As jumping to tasks is central to vit's operation, you might want to map each
```
thehunmonkgroup: Most of the rest of the doc has `VIT` capitalized, can you make that the same here?

CUSTOMIZE.md
```diff
@@ -110,3 +110,27 @@ class Keybinding(object):
...
+digit key to an ex command containing that digit, by adding the following to
```
thehunmonkgroup: Maybe `ex` to emphasize?

thehunmonkgroup (changes requested):

PR #278, opened 2020-12-24
Add 'abort_backspace' config, False by default.

### LINKED ISSUE #274: Quit command mode when pressing backspace in an empty command bar
**Is your feature request related to a problem? Please describe.**

Press `:`, press `<backspace>`. The colon should then disappear and vit return to normal mode.
This is how (neo)vim behaves.

**Describe the solution you'd like**

Add a conditional statement.

**Additional context**

None.

champignoom:
If true, hitting backspace against an empty prompt aborts the prompt.

Resolves https://github.com/vit-project/vit/issues/274

### DIFF
--- a/vit/command_bar.py
+++ b/vit/command_bar.py
@@ -5,13 +5,12 @@ class CommandBar(urwid.Edit):
     """Custom urwid.Edit class for the command bar.
     """
     def __init__(self, **kwargs):
-        self.event = kwargs['event']
-        self.autocomplete = kwargs['autocomplete']
+        self.event = kwargs.pop('event')
+        self.autocomplete = kwargs.pop('autocomplete')
+        self.abort_backspace = kwargs.pop('abort_backspace')
         self.metadata = None
         self.history = CommandBarHistory()
         self.readline = Readline(self)
-        kwargs.pop('event')
-        kwargs.pop('autocomplete')
         return super().__init__(**kwargs)
 
     def keypress(self, size, key):
@@ -20,15 +19,7 @@ class CommandBar(urwid.Edit):
         if key not in ('tab', 'shift tab'):
             self.autocomplete.deactivate()
         if 'choices' in self.metadata:
-            op = self.metadata['op']
-            data = {
-                'choice': None,
-                'metadata': self.get_metadata(),
-            }
-            if key in self.metadata['choices']:
-                data['choice'] = self.metadata['choices'][key]
-            self.cleanup(op)
-            self.event.emit('command-bar:keypress', data)
+            self.quit({'choice': self.metadata['choices'].get(key)})
             return None
         elif key in ('up',):
             self.readline.keypress('ctrl p')
@@ -38,16 +29,9 @@ class CommandBar(urwid.Edit):
             return None
         elif key in ('enter', 'esc'):
             text = self.get_edit_text().strip()
-            metadata = self.get_metadata()
-            data = {
-                'key': key,
-                'text': text,
-                'metadata': metadata,
-            }
-            self.cleanup(metadata['op'])
-            if text and key in ('enter'):
-                self.history.add(metadata['history'], text)
-            self.event.emit('command-bar:keypress', data)
+            if text and key == 'enter':
+                self.history.add(self.metadata['history'], text)
+            self.quit({'key': key, 'text': text})
             return None
         elif key in ('tab', 'shift tab'):
             if self.is_autocomplete_op():
@@ -59,7 +43,13 @@ class CommandBar(urwid.Edit):
             return None
         elif key in self.readline.keys():
             return self.readline.keypress(key)
+        elif self.is_aborting_backspace(key):
+            self.quit({'key': key})
+            return None
         return super().keypress(size, key)
+
+    def is_aborting_backspace(self, key):
+        return key == 'backspace' and self.abort_backspace and not self.get_edit_text()
 
     def is_autocomplete_op(self):
         return self.metadata['op'] not in ['search-forward', 'search-reverse']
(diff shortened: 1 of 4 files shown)


vit/command_bar.py
```diff
@@ -59,8 +45,20 @@ def keypress(self, size, key):
...
+        elif key in ('backspace',) and self.abort_backspace and not self.get_edit_text():
+            self.quit({'key': key})
+            return None
         return super().keypress(size, key)
 
+    def quit(self, metadata_args={}):
+        self.set_caption('')
+        self.set_edit_text('')
+        self.history.cleanup(self.metadata['op'])
+        data = {'metadata': self.get_metadata(), **metadata_args}
```
thehunmonkgroup: Instead of ditching the ```cleanup()``` method, let's keep that and call it after getting ```data``` in ```quit()``` -- that seems a better separation of concerns.

vit/command_bar.py
```diff
@@ -7,6 +7,7 @@ class CommandBar(urwid.Edit):
     def __init__(self, **kwargs):
         self.event = kwargs['event']
         self.autocomplete = kwargs['autocomplete']
+        self.abort_backspace = kwargs.pop('abort_backspace')
```
thehunmonkgroup, line 10: We're not using leftover ```kwargs```, so the ```pop()``` seems unnecessary or inconsistent here. We should either ```pop()``` all the ```kwargs``` or none of them.

vit/command_bar.py
```diff
@@ -59,8 +45,20 @@ def keypress(self, size, key):
             return None
         elif key in self.readline.keys():
             return self.readline.keypress(key)
+        elif key in ('backspace',) and self.abort_backspace and not self.get_edit_text():
```
thehunmonkgroup: I"m wondering if breaking this line off into a separate ```abort_on_backspace()``` method would be cleaner? I think it might make the code a bit more readable there.

vit/command_bar.py
```diff
@@ -20,15 +21,7 @@ def keypress(self, size, key):
...
-            op = self.metadata['op']
-            data = {
-                'choice': None,
-                'metadata': self.get_metadata(),
-            }
-            if key in self.metadata['choices']:
-                data['choice'] = self.metadata['choices'][key]
-            self.cleanup(op)
-            self.event.emit('command-bar:keypress', data)
+            self.quit({'choice': self.metadata['choices'].get(key)})
```
thehunmonkgroup, line 22: Nice use of the ```get()``` here to clean up the old code a bit! Note to self to use that more often...

thehunmonkgroup (changes requested):
Haven't tested it yet, but the code is clean and pretty close.

vit/command_bar.py
champignoom: The rationale is that the kwargs not used by the parent constructor, and only them, must be popped out. Otherwise an exception would be raised:

```
  File "command_bar.py", line 16, in __init__
   return super().__init__(**kwargs)
TypeError: __init__() got an unexpected keyword argument 'abort_backspace'
```

vit/command_bar.py
champignoom: It was like that before, but I later removed the `cleanup()` because it had caused an exception upon the next keypress when it is used as the only step of processing backspace, and thereby confused me what exactly it intends to do, considering that a "cleanup" would usually not corrupt what is cleaned up. If the separation is still preferred, I would suggest to document/rename it better.

vit/command_bar.py
champignoom: Agreed, but I would call it `is_aborting_backspace(key)` if the function contains all three conditions in the line you quoted.

vit/command_bar.py
thehunmonkgroup: Ah, yep, that makes sense. Maybe a code note in there to clarify why we're ```pop```ing it?

vit/command_bar.py
thehunmonkgroup: OK, reasonable, go ahead with that name you suggested.

vit/command_bar.py
thehunmonkgroup: Well, that method handles all of the tasks of restoring things to a pristine state, and that does seem like a rational grouping of tasks for a method. Do you have any suggestions on a more clear name?

champignoom: Maybe `deactivate`?

vit/command_bar.py
champignoom: Good idea. I'll do it.

.gitignore
```diff
@@ -1,9 +1,138 @@
-.python-version
```
thehunmonkgroup: I merged #279, so this needs to come back out

vit/command_bar.py
```diff
@@ -73,19 +63,23 @@ def set_edit_text(self, text, edit_pos=None):
...
 
     def activate(self, caption, metadata, edit_text=None):
         self.set_metadata(metadata)
         self.set_command_prompt(caption, edit_text)
 
-    def cleanup(self, command):
-        self.set_caption('')
-        self.set_edit_text('')
-        self.history.cleanup(command)
+    def deactivate(self):
```
thehunmonkgroup, line 73: Hm, this still doesn't feel quite right to me. The purpose of the method is to perform all actions necessary to reset the command bar to a 'fresh' state. How about something like ```reset_data``` or ```reset_configuration```?

thehunmonkgroup (changes requested):

.gitignore
champignoom: It's already there in the 'pyenv' section:
https://github.com/vit-project/vit/blob/272126da4e94a64946765df8173cd801fbee68e7/.gitignore#L88

vit/command_bar.py
champignoom: The name 'deactivate' was proposed becaused 1) the function undoes `activate()`, and 2) it signifies that calling the function turns the command bar into an unusable state, which is not apparent from 'reset' or 'cleanup'.

You have the final say.

thehunmonkgroup: You talked me into it, let's leave ```deactivate``` as the method name, at least for now.

.gitignore
thehunmonkgroup: Perhaps we're talking past each other. What I meant was that I already merged #279, which includes all these same changes in .gitignore. I think the correct approach given that is:
1. Remove the changes to .gitignore in this PR
2. Rebase this branch against 2.x
3. Force push this branch

champignoom: Sorry for the misunderstanding. The #279 is now excluded from this PR.

thehunmonkgroup (approved):

thehunmonkgroup:
OK, ran some tests and gave the code another look, this looks just about good to go.

Only remaining task is to squash the commits in this PR, let's get this in one changeset for clarity.

champignoom:
done

thehunmonkgroup:
@champignoom excellent work on this PR, thanks for your contribution!

Issue #240, opened 2020-06-05
Methods for modifying multiple tasks


jmriego:
Hi! First of all, thank you for your hard work. I found out very recently about vit and I find it amazing. I can't believe it's not a lot more popular between the Taskwarrior tools!

I was trying to make changes in multiple tasks at the same time and was wondering if there's some way of doing this.

The two ways I would think of are similar as what you would do in vim:
1. Mark multiple tasks with something like `mt` and then you run a command that would affect all of them. This should also work for when you have mappings that use the task UUID and would run them multiple times with a different UUID each
2. Add a `.` mapping that will repeat the previous action. In my opinion itt would also have to work for things like `:!rw task ....` so it would be a bit different of the same mapping in vim that don't include `:commands`.

Does this sounds like something that could be added? Thanks!

jmriego:
I suppose another option would be a `v` or `V` mapping to start a selection but that would require the tasks you want to modify to be next to each other

thehunmonkgroup:
Multi-task operations have been on my wish list for awhile.

It's not something I have time to implement now, but I'd be happy to look at a PR for it.

I do suspect the implementation will be a lot trickier than you may imagine. Adjusting the interface to display multiple selections and operate on multiple tasks will likely involve some fairly major refactors, and a poor implementation would likely be pretty buggy.

jmriego:
I can imagine! That's why I mentioned several methods that could potentially work. As you said, there's the interface part with also changes in the highlighting. You have to think about multiple types of changes and running task commands that you don't know what could affect...
So I think the easiest method would probably be the dot, but what do you think? I might have time to have a look at the code if that makes sense

thehunmonkgroup:
The dot would probably be a good start. I still think that might have some dragons in it, but conceptually, "store last action, repeat when told" seems fairly straightforward.

Also, I think even if we start here, adding the multiple highlighting later wouldn't conflict, it would just be another usable method.

thehunmonkgroup:
#267 is taking a first stab at this, still a long ways to go...

kevinstadler:
Since bulk-editing is also a feature I'd really like to be able to use from within vit I've had a little look at the code. Here's some ideas, would love to hear some feedback before I go forward and actually implement anything:

[...]

PR #289, opened 2021-03-28
Add support for {*_FILTER} keybinding replacements (refs #240)

kevinstadler:
This pull request adds support for 4 new keybinding replacement variables, each of which resolves to `task` filter expressions that underly the currently visible list of tasks displayed by vit:

* `{EXTRA_FILTER}`: any extra filter expression applied by the user
* `{REPORT_FILTER}`: the `report.___.filter` of the currently visible report
* `{CONTEXT_FILTER}`: filter expression of the currently active context
* `{VIEW_FILTER}`: this is just a concatenation of the other three which preserves internal precedence by adding parentheses (if necessary)

## Testing

The easiest way to test the correctness of the replacements is by adding the following keybinding (change the actual key to your liking): `G = :all {VIEW_FILTER}`
Now open any report in any context in vit and apply any extra filters. By invoking the binding you can first inspect the correctness of the filter expression, when you confirm it you will end up at the `all` report showing the exact same tasks as the view that you just came from (tasks shown and tasks completed in the status bar should be the same as well). Repeatedly invoking the binding will just bring you back to the same report with the same filter applied over and over again.

### DIFF
--- a/vit/application.py
+++ b/vit/application.py
@@ -199,10 +199,49 @@ class Application():
                     return str(task[attribute])
             return ''
 
+        # currently active filter lookup functions. all of these return the
+        # internally used filter representation, which is a flat list of strings
+        _extra_filter = lambda: self.extra_filters
+        _report_filter = lambda: self.model.active_report_filter()
+        _context_filter = lambda: self.active_context_filter()
+
+        # _view_filter is simply a concat of these three filters. logical 'and' is
+        # implied by taskwarrior, just need to make sure to preserve the internal
+        # precedence within each of the three sub-filters
+        def _protect_precedence(filter):
+            return ['(', *filter, ')'] if len(filter) > 1 else filter
+        def _view_filter():
+            parts = [ _extra_filter(), _report_filter(), _context_filter() ]
+            # filter out empty parts
+            parts = [ part for part in parts if part ]
+            # if there is more than one non-empty filter, and at least one part
+            # potentially has complex internal precedence structure...
+            if len(parts) > 1 and any([ len(part) > 1 for part in parts ]):
+                # ..then add parens around all complex parts
+                parts = [ _protect_precedence(part) for part in parts ]
+            return [ item for sublist in parts for item in sublist ]
+
+        _filter_replacements = {
+            'EXTRA_FILTER': _extra_filter,
+            'REPORT_FILTER': _report_filter,
+            'CONTEXT_FILTER': _context_filter,
+            'VIEW_FILTER': _view_filter,
+        }
+        def _filter_attribute_match(variable):
+            if variable in _filter_replacements:
+                return [variable]
+        def _filter_attribute_replace(task, variable):
+            # get filter resolution function and execute it, return as string
+            return ' '.join(_filter_replacements.get(variable)())
+
         replacements = [
             {
                 'match_callback': _task_attribute_match,
                 'replacement_callback': _task_attribute_replace,
+            },
+            {
+                'match_callback': _filter_attribute_match,
+                'replacement_callback': _filter_attribute_replace,
             },
         ]
         return replacements
--- a/vit/application.py
+++ b/vit/application.py
@@ -97,6 +97,9 @@ class Application():
     def set_active_context(self):
         self.context = self.task_config.get_active_context()
 
+    def active_context_filter(self):
+        return self.contexts[self.context]['filter'] if self.context else []
+
     def load_contexts(self):
         self.contexts = self.task_config.get_contexts()
 
@@ -195,6 +198,7 @@ class Application():
                 else:
                     return str(task[attribute])
             return ''
+
         replacements = [
             {
                 'match_callback': _task_attribute_match,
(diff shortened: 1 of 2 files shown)


vit/application.py
```diff
@@ -195,11 +198,51 @@ def _task_attribute_replace(task, attribute):
...
+        # precedence within each of the three sub-filters
+        def _protect_precedence(filter):
+            return ['(', *filter, ')'] if len(filter) > 1 else filter
+        def _view_filter():
+            parts = [ _extra_filter(), _report_filter(), _context_filter() ]
+            # filter out empty parts
+            parts = [ part for part in parts if part ]
+            # if there is more than one non-empty filter, and at least one part
+            # potentially has complex internal precedence structure...
+            if len(parts) > 1 and any([ len(part) > 1 for part in parts ]):
```
thehunmonkgroup, line 219: This seems like it might be overkill -- is there any reason why we can't always `_protect_precedence()`?

thehunmonkgroup:
Code looks good on initial review, nice work! Just that one small comment to address.

I don't really have time to do any extensive testing, can anyone else jump in here to make sure we're not breaking anything?

thehunmonkgroup:
One other thought: Most of the current keybinding replacements start with `TASK_`, which I like because the name goes general to specific. For these, I'm thinking we might want to flip them around, `FILTER_VIEW`, etc.

Thoughts?

vit/application.py
kevinstadler: The point is to avoid adding parentheses which would lead to really cluttered-looking replacements that are not necessary. E.g. if there is currently only one filter (say the report has `status:PENDING or +next`, no other filters) then the `{VIEW_FILTER}` would unnecessarily resolve to `( status:PENDING or +next )`, running the same binding again would lead to `( ( status:PENDING or +next ) )` and so on. The current solution is not perfect, but it does the job for 90% of cases with one inexpensive line of code.

I was thinking about making this a dedicated helper function as it might be useful in other situations as well, but wasn't sure about naming convention so didn't do it. How would you feel about having a method such as the following as part of the Application class, or is there a file where small static helper functions can go?
```python
# argument is a list of flat lists, returns one flat list expression which preserves the
# internal precedence that any of the filters might have
def merge_filters(filters):
  ...
```

thehunmonkgroup: `util.py` already exists, and seems like a reasonable location for it.

kevinstadler:
Very good point about naming. Funnily enough I read the existing `TASK_...` replacements as having a semantic `OBJECT_ATTRIBUTE` or `PARENT_CHILD` format, so I went for `..._FILTER` because in each case there is some task/vit object (the report, context or view) each of which has the same type of attribute (a filter expression).

I'm happy with whatever naming you think is better, but note that there is already a `ACTION_FILTER_REPORT` magic word, so I feel like having a `FILTER_REPORT` replacement could be confusing. Even if actions are marked with the `ACTION_` prefix I had the feeling that `FILTER_VIEW` suggests reading 'filter' as a verb, i.e. instructing vit to perform an action rather than return an attribute.

If you think a common prefix for all replacements would be good, how about adding `CURRENT_` or `ACTIVE_`? I.e. there would be `CURRENT_REPORT_FILTER`, `CURRENT_CONTEXT_FILTER` etc.

thehunmonkgroup:
I'm pretty sure it's `ACTION_REPORT_FILTER` and not `ACTION_FILTER_REPORT`.

I wouldn't be too concerned if these were internal names, but they are user-facing, and it seems like a good idea to pause and do our best to get them right.

Going with your understanding of object/attribute naming, I'll counter with these names:

`REPORT_FILTER`
`REPORT_FILTER_EXTRA`
`CONTEXT_FILTER`
`VIEW_FILTER`

So the 'objects' are report/context/view, and the attribute is filter.

Thoughts?

kevinstadler:
Ah yes you're right about the action name, and also true about getting the names right. The 'extra filters' are a weird one since it's not really clear what task/vit entity they belong to, I'm not sure they actually belong to the report but if the action that applies them is already called `_REPORT_FILTER` then it's only consistent to keep that name.

Do you think `VIEW_FILTER` is appropriate? In the code the object that actually displays all tasks in the end is the 'TaskList', but I don't think it has actually been referred to as anything in the documentation, so I thought that 'view' might be more intuitive.

If you're ok with the names I will probably create a new pull request (from a separate branch where the helper functions are neatly stowed away in `util.py`).

thehunmonkgroup:
The name I suggested for the extra works for me.

The view name is suspect, but I'm not 100% sure of a better name,  either.

Usually when I have trouble naming something, it's because it's not been clearly defined in my own mind enough yet.

So what is that? A union of all filters? A combination?

```
REPORT_FILTER_ALL
REPORT_FILTER_COMBINED
FILTER_ALL
FILTER_COMBINED
```

I dunno, none of those completely satisfy, just hoping they jog something in the conversation.

kevinstadler:
Yeah the 'view' is a weird one because it's in a way the most intuitive one of all (every task list you ever look at is based on applying it), but it doesn't have its own name, it's just...*the list*.

Tell you what, the keybinding replacement is actually just an optional stepping stone on the way to bulk editing. Having figured out the internal logic of the filters I think it might be more fruitful if I just have a go at the next stop of implementing bulk editing, without worrying about introducing new user-facing magic words (which would probably only see rather infrequent use anyway, especially once bulk editing proper is implemented).

thehunmonkgroup:
Yep, love it. Go for it.

Issue #307, opened 2021-10-10
Vit ignores new taskwarrior configuration 'report.X.context'


RonMcKay:
**Describe the bug**
In taskwarrior 2.6.0 there is a new configuration option for reports to ignore the currently set context `report.<name>.context=0`.
Vit seems to ignore this new setting.

**To Reproduce**
In a `.taskrc` put the following:
```
report.next.context=0
context.work.read=project:Work
context.work.write=project:Work
```

Then execute the following:
```
task context none
task add project:Work First Task
task add SecondTask
task context work
```
Then show the 'next' report in vit

**Expected behavior**
The report should show both tasks 'First Task' and 'Second Task'.
The 'next' report in vit however does only show the 'First Task'

**Test case**
I have attached the modified `generate-dummy-install.sh` which reproduces the issue (I had to change the filetype to `.txt` to be able to attach it to this issue).
[generate-dummy-install.txt](https://github.com/vit-project/vit/files/7318354/generate-dummy-install.txt)

In my setup I use the following vit and taskwarrior versions:
vit: Commit 0df4841
taskwarrior: 2.6.0

thehunmonkgroup:
May not have time to look at this for awhile, I would welcome a pull request to address the issue.

RonMcKay:
I created a pull request but as I do not have a complete overview of the code base I might have missed something.

PR #308, opened 2021-10-12
Fix 307 - Vit ignoring 'report.X.context=0' option

RonMcKay:
As mentioned in #307 Taskwarrior 2.6.0 introduced a new option 'report.X.context'. If set to 0 the respective report should ignore all filters set by the current context.

This PR tries to fix this missing feature within VIT.

I tested this fix with TW 2.6.0 and the dummy install attached to #307 but it should also be backwards compatible (please someone test this with an older TW version).

### DIFF
--- a/vit/application.py
+++ b/vit/application.py
@@ -922,7 +922,7 @@ class Application():
         self.task_config.get_projects()
         self.refresh_blocking_task_uuids()
         self.formatter.recalculate_due_datetimes()
-        context_filters = self.contexts[self.context]['filter'] if self.context else []
+        context_filters = self.contexts[self.context]['filter'] if self.context and self.reports[self.report].get('context', 1) else []
         try:
             self.model.update_report(self.report, context_filters=context_filters, extra_filters=self.extra_filters)
         except VitException as err:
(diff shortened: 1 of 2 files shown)


thehunmonkgroup:
I can't make sense of what's happening in the PR because there are a bunch of whitespace changes.

I"m guessing your editor has some kind of auto-whitespace correction? If so, you'll need to disable that and update the PR with only the actual code changes.

thehunmonkgroup:
I see in your PR's commit history that you have a commit fixing whitespace changes. Missed that the first time. Can you instead put those in a separate PR? It's cleaner to do those kind of fixes separately.

RonMcKay:
> 

> I see in your PR's commit history that you have a commit fixing whitespace changes. Missed that the first time. Can you instead put those in a separate PR? It's cleaner to do those kind of fixes separately.

Yeah, I saw that some code blocks had only 2 space indentations or were double indented. With the fix the code was more consistend with the rest of the codebase.
I removed the commit from the PR and force pushed the branch. I will then open a second PR as soon as this one is merged.

PR #315, opened 2021-10-26
Focus newly created task

arozans:
This PR tries to close the issue: https://github.com/vit-project/vit/issues/130

### DIFF
--- a/vit/application.py
+++ b/vit/application.py
@@ -356,6 +356,7 @@ class Application():
                     if self.execute_command(['task', 'add'] + args, wait=self.wait):
                         task = self.task_get_latest()
                         self.activate_message_bar('Task %s added' % task_id_or_uuid_short(task))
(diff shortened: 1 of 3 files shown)


thehunmonkgroup:
The code looks good.

One question: is `focus_added_task` the name we want for the config option? It's best to get this right the first time :) Is there any other name that's more succinct and/or clear?

tbabej:
Perhaps `focus_on_add`?

arozans:
I like @tbabej version too; I think we can also extend config comment with info that task must be visible on active filter to focus on it.

thehunmonkgroup:
Let's go with `focus_on_add`, and also please flesh out the config comment to describe the full behavior of the setting.

arozans:
@thehunmonkgroup can we merge this? :)

PR #323, opened 2022-03-22
Simplify timezone handling

jspricke:
astimezone() assumes the system local timezone if called without
arguments .

https://docs.python.org/3.10/library/datetime.html#datetime.datetime.astimezone

### DIFF
--- a/vit/formatter/__init__.py
+++ b/vit/formatter/__init__.py
@@ -153,14 +153,14 @@ class DateTime(Formatter):
     def age(self, dt):
         if dt == None:
             return ''
-        now = datetime.now(self.formatter.zone)
+        now = datetime.now().astimezone()
         seconds = (now - dt).total_seconds()
         return self.format_duration_vague(seconds)
 
     def countdown(self, dt):
         if dt == None:
             return ''
-        now = datetime.now(self.formatter.zone)
+        now = datetime.now().astimezone()
         if dt < now:
             return ''
         seconds = (dt - now).total_seconds()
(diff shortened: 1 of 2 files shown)


thehunmonkgroup:
While this reduces the amount of code slightly, the current code also seems more clear to me, as it requires no understanding of the default behavior of `astimezone()`, and also is not subject to the (admittedly unlikely) chance of a change in behavior of the `astimezone()` default.

Can you explain to me the advantage you see to this change?

jspricke:
* thehunmonkgroup ***@***.***> [2022-03-22 06:43]:
>Can you explain to me the advantage you see to this change?

ZoneInfo('localtime') needs a localtime file in zoneinfo.TZPATH which is 
not the case for a number of operating systems. Also the code only uses 
the zone to crate a timezone aware datetime to compare it to other 
datetimes and does not need a full featured ZoneInfo object.

thehunmonkgroup:
```
[:~/git/vit] jspricke-timezone+ ± vit
  File "vit/command_line.py", line 8, in <module>
    main()
  File "vit/command_line.py", line 5, in main
    Application(options, filters)
  File "/home/hunmonk/git/vit/vit/application.py", line 77, in __init__
    self.refresh(False)
  File "/home/hunmonk/git/vit/vit/application.py", line 921, in refresh
    self.build_main_widget()
  File "/home/hunmonk/git/vit/vit/application.py", line 969, in build_main_widget
    self.update_report(self.report)
  File "/home/hunmonk/git/vit/vit/application.py", line 944, in update_report
    self.update_task_table()
  File "/home/hunmonk/git/vit/vit/application.py", line 574, in update_task_table
    self.table.update_data(self.reports[self.report], self.model.tasks)
  File "/home/hunmonk/git/vit/vit/task_list.py", line 93, in update_data
    self.build_rows()
  File "/home/hunmonk/git/vit/vit/task_list.py", line 237, in build_rows
    formatted_value = column['formatter'].format(task[column['name']], task)
  File "/home/hunmonk/git/vit/vit/formatter/entry_age.py", line 5, in format
    return self.age(entry)
  File "/home/hunmonk/git/vit/vit/formatter/__init__.py", line 156, in age
    now = datetime.now(self.formatter.zone)
AttributeError: 'FormatterBase' object has no attribute 'zone'
```

jspricke:
Thanks for catching that, should be fixed.

PR #331, opened 2022-10-19
Allow flash configuration

### LINKED ISSUE #270: Allow disabling flash
**Is your feature request related to a problem? Please describe.**

When I use `vit` and perform an action like "start a task" and "stop a task" the task I'm working on flashes.

I believe the internal function used for this is `flash_focus`.

**Describe the solution you'd like**

Ideally, I would like to disable flashing completely via config option or similar.

**Additional context**

I'd like `vit` to be as fast as possible so I don't like waiting for flashes.

JKubovy:
Add 'flash_focus_repeat_times' and 'flash_focus_pause_seconds' keys to [vit] section in config.ini. Also support int and float types in config and automatically parse (transform) them. Keep same default values.

fixes vit-project/vit#270

### DIFF
--- a/vit/config_parser.py
+++ b/vit/config_parser.py
@@ -47,6 +47,8 @@ DEFAULTS = {
         'abort_backspace': False,
         'focus_on_add': False,
         'pid_dir': '',
+        'flash_focus_repeat_times': 2,
+        'flash_focus_pause_seconds': 0.1,
     },
     'report': {
         'default_report': 'next',
@@ -166,7 +168,7 @@ configuration.
         try:
             value = self.config.get(section, key)
             return self.transform(key, value, default)
-        except (configparser.NoSectionError, configparser.NoOptionError):
+        except (configparser.NoSectionError, configparser.NoOptionError, ValueError):
             return default
 
     def items(self, section):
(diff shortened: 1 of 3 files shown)


vit/config_parser.py
```diff
@@ -166,7 +168,7 @@ def get(self, section, key):
         try:
             value = self.config.get(section, key)
             return self.transform(key, value, default)
-        except (configparser.NoSectionError, configparser.NoOptionError):
+        except (configparser.NoSectionError, configparser.NoOptionError, ValueError):
```
thehunmonkgroup, line 171: Why did you add `ValueError` here?

JKubovy: Because if you parse int or float you can get ValueError. In that case you need to use default value.

thehunmonkgroup:
Works great, thanks for your efforts!

scottkosty:
@JKubovy Thanks! Just curious, which values do you choose for the new configs? (edit: ah, after looking at the issue I am guessing you use `flash_focus_repeat_times=0`)

thehunmonkgroup:
In my own configuration? The defaults, I'm happy with the way it works now.

PR #332, opened 2022-10-23
Minor fixes involving IndexErrors

JustBobinAround:
I noticed that when there are no tasks present, certain functions in application.py and base_list_box.py could crash the application. This pull requests fixes the issues by checking the lists' lengths are greater than 0.

### DIFF
--- a/vit/application.py
+++ b/vit/application.py
@@ -542,20 +542,21 @@ class Application():
         rows = self.table.rows
         current_index = start_index
         last_index = len(rows) - 1
-        start_matches = self.search_row_has_search_term(rows[start_index], search_regex)
-        current_index = self.search_increment_index(current_index, reverse)
-        while True:
-            if reverse and current_index < 0:
-                self.search_loop_warning('TOP', reverse)
-                current_index = last_index
-            elif not reverse and current_index > last_index:
-                self.search_loop_warning('BOTTOM', reverse)
-                current_index = 0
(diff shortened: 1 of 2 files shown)


vit/application.py
```diff
@@ -542,20 +542,21 @@ def search_rows(self, term, start_index=0, reverse=False):
...
-                self.search_loop_warning('TOP', reverse)
-                current_index = last_index
-            elif not reverse and current_index > last_index:
-                self.search_loop_warning('BOTTOM', reverse)
-                current_index = 0
-            if self.search_row_has_search_term(rows[current_index], search_regex):
-                return current_index
-            if current_index == start_index:
-                return start_index if start_matches else None
+        if last_index > -1:
```
thehunmonkgroup: This test isn't the most intuitive. I think it would be more clear to check `len(rows)` directly?

thehunmonkgroup (changes requested):

vit/application.py
JustBobinAround: I agree, bad judgment on my part. That would match the other fixes better too.

thehunmonkgroup:
Looks good now. Can you squash your work into one commit, then I think it will be ready for merge.

JustBobinAround:
The commits have been squashed.

thehunmonkgroup:
Thanks for your work!

PR #335, opened 2023-03-01
A few warning fixes

mattst88:


CUSTOMIZE.md
```diff
@@ -94,7 +94,7 @@ Here's a minimal example:
 
 ```python
 # keybinding/keybinding.py
-class Keybinding(object):
+class Keybinding():
```
thehunmonkgroup: https://docs.python.org/3/tutorial/classes.html#class-definition-syntax seems to suggest we go even further and do `class Keybinding:`

Perhaps we should adjust the rest of these to match that syntax?

thehunmonkgroup (changes requested):
Overall these seem like easy and sensible changes, and should be easy to get in once you address the one question listed in the review.

CUSTOMIZE.md
mattst88: Ah, good point. Yes, I'll make that change.

thehunmonkgroup:
Thanks for your work on this!

mattst88:
Thank you for the awesome software!

PR #346, opened 2023-09-17
fix(py3.12): `SafeConfigParser` -> `ConfigParser`

sanjayankur31:
`SafeConfigParser` was deprecated in 3.2 and has been removed.

Reference:

https://docs.python.org/3.12/whatsnew/3.12.html#configparser


thehunmonkgroup:
@sanjayankur31 thanks for your contribution!

PR #357, opened 2024-07-06
Add script to reload events if pending.data changed

jspricke:
This needs the Python watchdog library

### DIFF
--- a/scripts/vit-watchdog
+++ b/scripts/vit-watchdog
@@ -1,0 +1,43 @@
+#!/usr/bin/env python3
+#
+# Watchdog script for vit -- automatically reloads vit when task data changes.
+# Monitors the pending.data file and sends a SIGUSR1 signal to vit to trigger a refresh.
+# Only works with Taskwarrior < 3.0 for now.
+
+from os.path import expanduser
+from subprocess import Popen, signal
+
+from vit.config_parser import ConfigParser, TaskParser
+from vit.loader import Loader
+
+try:
+    from watchdog.events import PatternMatchingEventHandler
+    from watchdog.observers import Observer
+except ModuleNotFoundError:
+    print("Please install the Python watchdog library")
+    exit(1)
+
+p = None
+
+
+class VitEventHandler(PatternMatchingEventHandler):
+    def __init__(self):
+        super().__init__(patterns=["*/pending.data"], ignore_directories=True)
+
+    def on_modified(self, event):
+        if p:
+            p.send_signal(signal.SIGUSR1)
+
+
+loader = Loader()
+config = ConfigParser(loader)
+task_config = TaskParser(config)
+data_location = task_config.subtree("data.location")
+data_location = expanduser(data_location)
+
+observer = Observer()
+observer.schedule(VitEventHandler(), data_location)
+observer.start()
+
+p = Popen("vit")
+p.wait()


scripts/vit-watchdog
```diff
@@ -0,0 +1,26 @@
...
+    def __init__(self):
+        super().__init__(patterns=["*/pending.data"], ignore_directories=True)
+
+    def on_modified(self, event):
+        if p:
+            p.send_signal(signal.SIGUSR1)
+
+
+observer = Observer()
+observer.schedule(VitEventHandler(), expanduser("~/.local/share/task"))
```
thehunmonkgroup: Would be better if this was not hard-coded, not all people store their task data here.

scripts/vit-watchdog
```diff
@@ -0,0 +1,26 @@
...
+observer.start()
+
+p = Popen("/usr/bin/vit")
```
thehunmonkgroup: Same with this, not everyone has their vit executable here.

scripts/vit-watchdog
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env python3
+
+from os.path import expanduser
+from subprocess import Popen, signal
+
+from watchdog.events import PatternMatchingEventHandler
```
thehunmonkgroup: Perhaps if watchdog is required, and is not a standard module available in a vit environment, that there should be some graceful failure here, to inform the user to install the package?

thehunmonkgroup (changes requested):

scripts/vit-watchdog
jspricke: done

scripts/vit-watchdog
jspricke: done

scripts/vit-watchdog
jspricke: done

thehunmonkgroup:
Changes look good. I think it would be helpful to add some documentation, what are your thoughts about the best location? Could be added as a comment at the top of the script, or in some of the other existing markdown doc in the repo?

scottkosty:
Note that Taskwarrior 3.0 (released in March) does not support pending.data. It now uses a single database (taskchampion.sqlite3) rather than separate databases for pending, completed, etc.

thehunmonkgroup:
> Note that Taskwarrior 3.0 (released in March) does not support pending.data

Yep, I saw this. My understanding of the upgrade is that VIT should work seamlessly with 3.x, as the CLI interface has not changed, only the storage mechanism.

I have not personally upgraded to 3.x, and will not do so until it's matured a bit more. I suspect I am not alone in this decision.

That being said, if SQLite is being used as the local storage mechanism, I would think this kind of file listener approach could still work?

jspricke:
> Could be added as a comment at the top of the script, or in some of the other existing markdown doc in the repo?

Added a comment.

> Yep, I saw this. My understanding of the upgrade is that VIT should work seamlessly with 3.x, as the CLI interface has not changed, only the storage mechanism.

Seems to work fine.

> I have not personally upgraded to 3.x, and will not do so until it's matured a bit more. I suspect I am not alone in this decision.

Same here. We also have to think about what to do with the Debian package as it will be a bit of work to get all that Rust stuff in.

> That being said, if SQLite is being used as the local storage mechanism, I would think this kind of file listener approach could still work?

Sadly it does not work as running `task` creates a write event to the database. @scottkosty do you know if that write event can be dropped?

scripts/vit-watchdog
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env python3
+#
+# Simple wrapper to reload vit upon task changes
```
thehunmonkgroup: The description feels a little vague.

Perhaps something more like this?

```python
# Watchdog script for vit -- automatically reloads vit when task data changes.
# Monitors the pending.data file and sends a SIGUSR1 signal to vit to trigger a refresh.
```

thehunmonkgroup:
> Sadly it does not work as running `task` creates a write event to the database.

I understand the purpose of this PR, as a simple solution to handle automatic updates, which is why I'll commit it when it's ready. However, there's already a documented method for doing auto-refresh: https://github.com/vit-project/vit/blob/2.x/CUSTOMIZE.md#auto-refreshing-vits-interface -- this method will continue to work in Taskwarrior 3.x

scripts/vit-watchdog
thehunmonkgroup: Also, it just occurred to me that the comment should mention that this will ONLY work with Taskwarrior < 3.0

scottkosty:
> > That being said, if SQLite is being used as the local storage mechanism, I would think this kind of file listener approach could still work?
> 
> Sadly it does not work as running `task` creates a write event to the database. @scottkosty do you know if that write event can be dropped?

Unfortunately I don't know anything about 3.x under the hood.

scripts/vit-watchdog
jspricke: done.

scripts/vit-watchdog
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env python3
+#
+# Watchdog script for vit -- automatically reloads vit when task data changes.
+# Monitors the pending.data file and sends a SIGUSR1 signal to vit to trigger a refresh.
```
thehunmonkgroup, line 4: I mentioned this before, we definitely want to add a note making it clear this script ONLY works with Taskwarrior < 3.0.

thehunmonkgroup (changes requested):

scripts/vit-watchdog
jspricke: done.

Issue #228, opened 2020-03-01
Sort is done alphabetically instead of in UDA value order from config


nfraprado:
When showing a report that is sorted by an UDA, VIT sorts the UDA values alphabetically, while Taskwarrior sorts using the order the UDA values are defined in the config.

To reproduce, use the following taskrc config:
```
data.location=~/test/data
uda.test.type = string
uda.test.label = Test
uda.test.values = a,b,c,z,y,x

report.test.columns = id,test,description
report.test.labels = ID,Test,Description
report.test.description = Test
report.test.sort = test-
```

Add tasks with each of the Test UDA values.

Display the Test report on VIT and Taskwarrior.

The sorting of the tasks will be, from top to bottom, based on the Test UDA values:

Taskwarrior: a,b,c,z,y,x
VIT : z,y,x,c,b,a

thehunmonkgroup:
This seems more like a bug in TaskWarrior.

```test-``` should sort descending, and the TaskWarrior result is ascending. Also, I've seen no documentation in TaskWarrior that UDA values should be sorted according to the order they are specified. Doesn't alphabetical make more sense anyways?

Finally, *all* of the sorting in VIT is structured around alphabetical sorting, so even if none of the above is true, it might be a real pain to treat UDA fields separately.

fdela:
IMHO, it rather makes sense to use the lexicographic order when sorting a UDA
* of type _string_
* _which has a specified set of values_ 
(if we don't care about order, we wouldn't have a rc.report._foo_.sort specify _field_[+-] anyway, so point would be moot).

It can be e.g. used with default priority field.
As I use it, I have configured uda.priority.values=H,M,,L so that tasks with Low priority are shown later than those with none specified when sorted descending.

I analyzed how it works in TaskWarrior: the entries are specified in _descending_ order, e.g.

Say we have (for a _string_ UDA) uda._field_.values = High, Medium, Low
Then we have tasks sorted as follows, according to the value of _field_:

rc.report._foo_.sort=_field_- => \<empty>, High, Medium, Low
rc.report._foo_.sort=_field_+ => Low, Medium, High, \<empty>
(\<empty> field value is arbitrarily considered the biggest value if not specified explicitely [as with my priority example hereabove, note the empty value is between the two consecutive commas]).

For _numeric_, _duration_ and _date_ types, uda._field_.values is unused, and natural order is used, so no change needed
[...]

PR #366, opened 2025-04-09
Fixed UDA & priority orders

riodelphino:
Consider priority and UDA orders in taskrc, such as `uda.priority.values = H,M,L,`.
Note that this only works if the UDA type is set to string.

taskrc:
```
uda.priority.values = H,M,L,
# uda.priority.type = string

report.list.sort = priority+
```

Before:
before
H -> L -> M (Wrong... Alphabetically...)

After:
after
H -> M -> L (Correct !)

This is my first pull request on GitHub.
Please let me know if there are any issues or if I've made any mistakes.

### DIFF
--- a/vit/task_list.py
+++ b/vit/task_list.py
@@ -177,7 +177,7 @@ class TaskTable:
     def sort(self):
         if 'sort' in self.report:
             for column, order, collate in reversed(self.report['sort']):
-                def comparator(first, second):
+                def comparator_default(first, second):
                     if first[column] is not None and second[column] is not None:
                         return -1 if first[column] < second[column] else 1 if first[column] > second[column] else 0
                     elif first[column] is None and second[column] is None:
@@ -186,6 +186,20 @@ class TaskTable:
                         return -1
                     elif first[column] is None and second[column] is not None:
                         return 1
+
+                def comparator_uda_string(values_index, first, second):
+                    a, b = first[column], second[column]
+                    default_index = len(values_index)
+                    a_index = values_index.get(a or '', default_index)
+                    b_index = values_index.get(b or '', default_index)
+                    return (a_index > b_index) - (a_index < b_index)
+
+                uda = self.task_config.uda_config.get(column)
+                if uda != None and uda.get('type') == 'string':
+                    values_index = uda.get('values_index')
+                    comparator = partial(comparator_uda_string, values_index)
+                else:
+                    comparator = comparator_default
                 if order and order == 'descending':
                     self.tasks = sorted(self.tasks, key=cmp_to_key(comparator), reverse=True)
                 else:
--- a/vit/task_list.py
+++ b/vit/task_list.py
@@ -195,7 +195,7 @@ class TaskTable:
                     return (a_index > b_index) - (a_index < b_index)
 
                 uda = self.task_config.uda_config.get(column)
-                if uda != None and uda.get('type') == 'string':
+                if uda is not None and uda.get('type') == 'string':
                     values_index = uda.get('values_index')
                     comparator = partial(comparator_uda_string, values_index)
                 else:


riodelphino:
#228

thehunmonkgroup:
Thanks for the PR. I will tell you up front that we have to be very careful with the sorting code, in particular because it's a real performance bottleneck for large reports -- some of my feedback will be addressing that.

1. Is there a more efficient way to test if a column is a UDA at all? As opposed to pulling and parsing data from all columns in case they are UDA
2. Do integer UDAs already sort correctly? I'm guessing yes. If so, then this patch breaks them.
3. I think it's a bad idea to always call `sorted()`, the existing code only calls it in the descending order case, for efficiency in the ascending case.

riodelphino:
@thehunmonkgroup Thank you for your quick reply.
I'm sorry — I'm just an amateur, and I submitted the PR without giving it enough thought.
I really appreciate your valuable feedback and will take some time to review my code carefully.
Please give me a little time.

riodelphino:
I've updated the code based on your suggestions, in commit [95863fa](https://github.com/vit-project/vit/commit/95863fad46d418a32450697c2ad9e5cfb0275c82)


## Updated
> 1. Is there a more efficient way to test if a column is a UDA at all? As opposed to pulling and parsing data from all columns in case they are UDA

- Cached UDA configs at the top of the `sort()` function.
- Added a `DEFAULT_COLUMNS` set to check whether a column is a UDA.

> 2. Do integer UDAs already sort correctly? I'm guessing yes. If so, then this patch breaks them.

- Now clearly bypasses numeric, date, and duration types to the default comparator function.
- Also bypasses string UDA which have no `.values`.

> 3. I think it's a bad idea to always call `sorted()`, the existing code only calls it in the descending order case, for efficiency in the ascending case.

- That was my mistake. It's been reverted.

## Testing
For the performance concern, I tested with the following settings:
```
uda.test.values = a,b,c,
uda.test.type = string
```
and used a report like:
```
report.all.sort = test-
```
With around 250 tasks, I didn't notice significant difference in rendering speed.
(Is it too few ?)

---

I'm not entirely sure if I’ve fully understood your suggestions, so please let me know if there are still any issues with my approach.

riodelphino:
I've made a small refinement to the sort logic in [19a59fa](https://github.com/vit-project/vit/commit/19a59fa6410df26ef5b6927dcd56799e81aaaed2)
This commit fixes a case where 'L' and empty values ('') were not sorted correctly.

thehunmonkgroup:
We're heading in a bit of the wrong direction here. We definitely do not want to keep a hard-coded list of non-UDA values around, that's brittle.

In fact, we shouldn't be parsing the UDA config in this code at all, that should be the job of `TaskParser` proper -- it can be done once at startup or reload. Also, `TaskTable` is already passed an instance of of the parsed task configuration in `task_config` when it's initialized -- there's no need to recreate that.

I've added https://github.com/vit-project/vit/commit/6625681803b40c89c8dc80963002f6f857684b77 to handle parsing UDA values as part of `TaskParser`, check that to see how to inspect the new `uda_config` attribute on the instance.

Also in https://github.com/vit-project/vit/commit/921f9fd89924f1652e19628eddad9a5c9e747737 I fixed a bug you would have introduced by stripping trailing commas from string values. Those are *valid* values, the value being an empty string.

So refactor your approach to inspect `self.task_config.uda_config` directly, I believe checking for a non-empty value of `values_index` would be sufficient to detect if your custom comparator should be used.

thehunmonkgroup:
Here's a dummy install script that can be used to test the performance of your change:

[custom-sort-dummy-install.sh.txt](https://github.com/user-attachments/files/19853704/custom-sort-dummy-install.sh.txt)

Make that executable and run it, and it will create about 4200 tasks in a temporary install, and give you instructions about how to execute VIT to access it, with this config:

```dosini
data.location=$TASK_DIR

uda.test.values = one,two,three,
uda.test.type = string

# Report sorted by priority ascending
report.priorityasc.description=Tasks sorted by priority (Ascending)
report.priorityasc.columns=id,priority,project,due,description
report.priorityasc.filter=priority.not:
report.priorityasc.sort=priority+,id+

# Report sorted by priority descending
report.prioritydesc.description=Tasks sorted by priority (Descending)
report.prioritydesc.columns=id,priority,project,due,description
report.prioritydesc.filter=priority.not:
report.prioritydesc.sort=priority-,id+

# Report sorted by test ascending
report.testasc.description=Tasks sorted by test (Ascending)
report.testasc.columns=id,test,project,due,description
report.testasc.filter=test.not:
report.testasc.sort=test+,id+

# Report sorted by test descending
report.testdesc.description=Tasks sorted by test (Descending)
report.testdesc.columns=id,test,project,due,description
report.testdesc.filter=test.not:
report.testdesc.sort=test-,id+
```

Once you get your PR sorted, I'd recommend running each of those reports several times with and without your change to check on the performance hit, and well as standard reports which don't use the custom comparator. That will allow us some confidence that the change will not negatively impact performance. Really, it's most important that it doesn't impact performance when sorting on columns other than UDA string type, as I think it's reasonable to give up a little performance on the string type sorting in exchange for it actually sorting correctly. ;)

riodelphino:
Thank you for the detailed explanation and for even updating the main branch to support my changes — I really appreciate it!
I'll try it again based on your new commits and run the performance tests.
I'm going to need a bit more time to revise the PR.

riodelphino:
I noticed that you wrote that shell script ! Thank you so much !
I was busy, and also wasn't sure whether to merge or rebase upstream/2.x, so it took me a while.

- Merged the latest `upstream/2.x` branch and reset my previous modifications.
- Then reapplied my UDA-related fixes in `task_list.py`.
- I inserted 4,200 dummy tasks using `custom-sort-dummy-install.sh`.
- export VIT_DIR and TASKRC
- Finally, I measured the report switching/display time for each report.

### Results:
#### upstream/2.x
priorityasc, prioritydesc, testasc, testdesc: around 2 seconds
all, list, long: around 2.8 seconds

#### My fork
priorityasc, prioritydesc, testasc, testdesc: around 2 seconds
all, list, long: around 2.8 seconds

I measured the time manually using the timer app on my iPhone ^^;
There doesn't seem to be any noticeable performance difference.

Please let me know if there are still any concerns or improvements needed.

thehunmonkgroup:
Something about the PR seems off, I'm seeing changes listed in `vit/config_parser.py` that have already been applied to the 2.x branch.

riodelphino:
I'm sorry for the trouble. I'm still not fully familiar with how GitHub and forks work...

I suspect that the issue may have been caused by accidentally running `git merge` in the `Reset to upstream/2.x` commit.

Is the following the correct approach for me to take?

```bash
git reset --hard HEAD~2
git rebase upstream/2.x
# git add & commit
# Add my fix again
# git add & commit
```

Or, should I make new PR with another branch created with current `upstream/2.x` ?

thehunmonkgroup:
i don't think you need to reset hard. if you have the latest 2.x code locally, a `git rebase -i` should allow you to see how the rebasing will work, squash/fixup extra commits, etc.

thehunmonkgroup:
after the `rebase -i` you may need to force push the changes: `git push -f`

riodelphino:
Thank you.

I tried `git rebase -i upstream/2.x` and squash `Reapply fixes` into `Reset to upstream/2.x`, and rename the commit as `Reapply fixes`.

But I'm not sure my procedure is right...

thehunmonkgroup:
The code looks the same, see here:



Another way to approach this is to do a hard reset to a commit before your work started, then merge 2.x to get to the tip of that branch, then manually apply your changes and commit them, then do a force push with `git push -f` -- clunky but it should work.

Most importantly, look at the 2.x branch, it already contains the code in `vit/config_parser.py`, so you shouldn't be including that in your PR.

riodelphino:
Thank you so much.
I reset to the commit where I originally started, rebased onto `upstream/2.x`, and then reapplied my fixes.


```bash
git reset --hard HEAD@{73}
git fetch upstream # again
git rebase upstream/2.x # apply your 2 commits
# Reapply my fixes here
```

I believe my commit [6b73751](https://github.com/vit-project/vit/pull/366/commits/6b7375156fe884fa73d5c5302b8cc3dcbb6575de) no longer includes your changes.
Please let me know if there's still anything I missed.

vit/task_list.py
```diff
@@ -186,6 +186,20 @@ def comparator(first, second):
...
+
+                def comparator_uda_string(values_index, first, second):
+                    a, b = first[column], second[column]
+                    default_index = len(values_index)
+                    a_index = values_index.get(a or '', default_index)
+                    b_index = values_index.get(b or '', default_index)
+                    return (a_index > b_index) - (a_index < b_index)
+
+                uda = self.task_config.uda_config.get(column)
+                if uda != None and uda.get('type') == 'string':
```
thehunmonkgroup: `uda is not None` is more idiomatic Python, let's use that.

riodelphino: Sure, thank you! I've updated the code.
Please take a look.

thehunmonkgroup:
@riodelphino all set, great work!

riodelphino:
Thank you so much for the long-term support and for merging!

I was too focused on not deleting my existing commits, thinking it was important for preserving the context of our discussion.
Now I understand better how complex commit operations are displayed and handled on GitHub.

I'm really glad the order of H → M → L could be preserved --- it was essential to me.

Thanks again !

